Issue #127, opened 2023-06-01
Support arbitrary key/value pairs as report metadata


epompeii:
Currently there is no way to add free form metadata to a report. Add support for arbitrary key/value pairs to be provided.
In the UI, display these metadata with the results.

epompeii:
Have key/value pairs could also be useful for the dimensions themselves (branch, testbed, benchmark).

Poolifier has very high dimensionality within their benchmarks: https://bencher.dev/perf/poolifier
Being able to easily filter and compare based on tag could be very useful here as well, especially in the Perf Plot.

jschwe:
As I mentioned in #539, this would be incredibly useful to track some metadata which could also be responsible for regressions or improvements of metrics, e.g.

- kernel / OS version (performance)
- compiler version (filesize, performance)

PR #146, opened 2023-06-14
Add (perf_data) upload to CLI report subcommand

brainstorm:
As discussed offline, adding a CLI subcommand to submit [firefox's profiler-server data](https://github.com/firefox-devtools/profiler-server/blob/master/API.md), example viz [here](https://profiler.firefox.com/public/6pw4z1nc9yd1k45qnzaevgc3zd1t5pzcp7ys3y8/calltree/?globalTrackOrder=0wd&thread=3&v=9).

Work in progress, I was unsure about:

1. In which CLI subcommand to drop this functionality, I hope `report` is a good place? The other candidate was `perf`, happy to move it there if you think it's more appropriate?
2. Should I POST a report [from the CLI](https://github.com/bencherdev/bencher/compare/main...brainstorm:bencher.dev:upload_perfdata?expand=1#diff-4179affbd8ea3ac6b16a686d63aeb440f683874fc5658236c34a98fb578da951R45-R50) and then have the bencher backend forward the POST of the received data to `localhost:5252/compressed-data`? OR have the CLI directly post it to the profile-server endpoint? I suspect that you want the former for several reasons... OTOH, the forwarding shuffles quite a few bytes on a couple of hops unless we send a 302 Redirect?

Anyway, let me know in advance if there's something off in my thinking or implementation so that we are on the same page, @epompeii.

Depends on https://github.com/bencherdev/bencher/issues/127


brainstorm:
Thinking about point 2 above... you wanted to host the firefox `profile-server` at bencher.dev or were you planning use the public instance instead (https://api.profiler.firefox.com/)?... ahh, re-reading your email it seems that you want the latter, so that means having a direct POST of the bencher API to that endpoint? Easy!

epompeii:
Thank you for taking the initiative on this, @brainstorm!

Yes, for the first pass lets just use the public instance (https://api.profiler.firefox.com). Long term, all this functionality should be built into the Bencher instead of adding an external dependency.

I was thinking that this wouldn't be a sub-command. Rather it would be an optional argument to `bencher run` (which is also aliased to `bencher report create`). Something along the lines of `--profile` that would be an `Option<PathBuf>`.
In the `bencher run` code, this would read the profile from the specified path _after_ all benchmarks have been run but _before_ the report has been posted to the API backend.
Basically right here: https://github.com/bencherdev/bencher/blob/main/services/cli/src/bencher/sub/project/run/mod.rs#L146

To keep this from being hard blocked on #127 you can just do all the work here and store it as `let _profile = ...` for now. Once I get the metadata added to the report, I'll just drop it in there!

epompeii:
@brainstorm I'm going to go ahead and close this out. Please, feel free to contribute towards https://github.com/bencherdev/bencher/issues/127 or reopen this issue once that work is complete.

PR #289, opened 2024-01-14
Remove forced `mold` linker on linux

Serial-ATA:
This blocked me from building, as it expects `mold` to be present *and* for it to be installed at `/usr/local/bin/mold`.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,10 +1,2 @@
 [alias]
 xtask = "run --manifest-path ./xtask/Cargo.toml --"
-
-[target.x86_64-unknown-linux-gnu]
-linker = "clang"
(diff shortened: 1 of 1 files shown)


epompeii:
I'm not sold on removing `mold` as a dependency for Linux builds, to help keep CI times in check.
I very much understand that it can be a temporary blocker when first getting started outside of a devcontainer though.
You're totally right though that the absolute path is overkill. I've change `/usr/local/bin/mold` to just `mold`: https://github.com/bencherdev/bencher/commit/f6af43d49e2f5c4cf0cc804f0fdfe58cbeeb9a32

Does that work for you?

Serial-ATA:
Sure that works. I already had mold installed, just not at that path.

PR #290, opened 2024-01-14
Remove user-specific settings pages

Serial-ATA:
This removes the user from the settings page URL.

https://bencher.dev/console/users/serial-ata/settings is now https://bencher.dev/console/settings

This makes it possible to link to the settings page(s) in a generic manner, for documentation purposes. (Ex. Create a new token at https://bencher.dev/console/settings/tokens)

I haven't done the organization settings yet, wanted to get an opinion on this change first.


epompeii:
Hey @Serial-ATA thanks for the interest in Bencher and putting thought towards this!

The `/console/users/serial-ata/settings` routes need to exist for admin purposes. Likewise the `/console/organizations/serial-ata/settings` pages are needed for the same reason.

I could see the a `/console/users/settings` route that is basically just a smart redirect to `/console/users/serial-ata/settings`.
Likewise we could have `/console/organizations/settings` be a smart redirect to `/console/organizations/serial-ata/settings`, if and only if the user only belongs to one organization. A user can belong to several organizations. If so, just redirect to `/console/organizations`.

This would likely work like similar to `/console`:
- https://github.com/bencherdev/bencher/blob/main/services/console/src/pages/console/index.astro
- https://github.com/bencherdev/bencher/blob/main/services/console/src/components/console/redirects/ConsoleRedirect.tsx

Serial-ATA:
Thanks for the quick reply. I'll look into getting that done soon.

Another thing I was thinking about doing is removing the `help` section from settings.
(https://bencher.dev/console/users/serial-ata/help) -> (https://bencher.dev/help)

Since it doesn't have anything user specific, besides the username which isn't necessary.

epompeii:
So `https://bencher.dev/help` already redirects when logged in to `https://bencher.dev/console/users/serial-ata/help`, so it is another example of the "smart redirect" I was talking about.
However, it doesn't need to be at `https://bencher.dev/console/users/serial-ata/help`. I would be okay with moving it to `https://bencher.dev/console/help`.
I want to keep it under `/console` though and separate from the public `/help` page.
Long term, I want the `/console/help` page to be very user/account specific and helpful. But all of that is TBD.

Yes, please target `devel`!
However, I would prefer for the default branch to remain as `main`. Due to the Bencher CLI GitHub Action living in this repo, it would be very confusing for folks to have the WIP branch be the first thing that they see.

I really appreciate all of the interest!
If you want to hop on a call to discuss the roadmap and such, feel free to grab a time here 😃
https://calendly.com/epompeii/call

Serial-ATA:
I'll close this for now, since I don't have much time to work on it right now. I'll return to this when I can (if it doesn't get done before then).

PR #319, opened 2024-02-08
Add Iai-Callgrind adapter

JonasWanke:
We're changing our continuous benchmarking in GitHub Actions (for our programming language [Candy](http://github.com/candy-lang/candy), implemented in Rust) to be based on Valgrind and discovered both Bencher and [Iai](https://github.com/bheisler/iai). However, Iai isn't maintained anymore (the last commit was three years ago), so we're using [Iai-Callgrind](https://github.com/iai-callgrind/iai-callgrind), which has a slightly different output format. Hence, I added an adapter supporting this format.

Iai-Callgrind exposes mostly the same measures as Iai:

| Iai              | Iai-Callgrind    |
| :--------------- | :--------------- |
| Instructions     | Instructions     |
| L1 Accesses      | L1 Hits          |
| L2 Accesses      | L2 Hits          |
| RAM Accesses     | RAM Hits         |
| —                | Total read+write |
| Estimated Cycles | Estimated Cycles |

> Their definitions/calculations seem to be the same:
>
> Iai: https://github.com/bheisler/iai/blob/58e28e940c62898a45d17cb531f9012fb173d236/src/lib.rs#L210-L239
> Iai-Callgrind: https://github.com/iai-callgrind/iai-callgrind/blob/main/iai-callgrind-runner/src/runner/callgrind/mod.rs#L129-L164

Therefore, I created a separate enum `IaiCallgrindMeasure` but reused the name/slug/unit strings where possible.

In the documentation, I only updated the English pages and not the “Benchmarking in Rust” section under `/learn`.

Also, I discovered a broken link in the documentation (pointing to localhost) which is now fixed.


lib/bencher_adapter/src/adapters/rust/iai.rs
```diff
@@ -224,7 +224,7 @@ pub(crate) mod test_rust_iai {
     }
 
     #[test]
-    fn test_adapter_rust_aia() {
```
epompeii, line 227: Good catch! 🙃

services/console/src/content/how_to/de/github-actions.mdx
```diff
@@ -42,7 +42,7 @@ Wenn Sie nur erwarten, Pull-Anfragen von Branches innerhalb des **gleichen** Rep
...
 2. Führen Sie bei `pull_request` Ereignissen aus, wenn und nur wenn der Pull Request aus dem gleichen Repository stammt. Für die Verarbeitung von Fork PRs siehe unten [Pull Requests von Forks](#pull-requests-from-forks).
 3. Legen Sie [die Berechtigungen für das `GITHUB_TOKEN`](https://docs.github.com/de/actions/using-jobs/assigning-permissions-to-jobs#setting-the-github_token-permissions-for-a-specific-job) auf `write` für `pull-requests`. Abhängig von Ihren GitHub-Einstellungen ist dies möglicherweise nicht erforderlich. Aber für alle Organisationen und persönlichen Repositories, [die nach dem 02. Februar 2023 erstellt wurden](https://github.blog/changelog/2023-02-02-github-actions-updating-the-default-github_token-permissions-to-read-only/), ist dies das Standardverhalten. Siehe [die GitHub-Dokumentation](https://docs.github.com/de/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token) für eine vollständige Übersicht.
-3. [Wählen Sie den Branch](http://localhost:3000/de/docs/explanation/branch-selection/) aus, den Sie verwenden möchten:
```
epompeii, line 45: Great catch! Thank you.

epompeii (approved):
Fantastic work, @JonasWanke ! Thank you for this great PR!
I've been a bit worried about `Iai` being abandonware, so it's great to learn that `iai-callgrind` exits 😃

I've changed the target branch to `devel`. I should have `v0.4.2` out in the next day or two with your changes included! 🚀

Issue #417, opened 2024-06-12
Pass in environment variables to `bencher up`


epompeii:
Currently `bencher up` does not support passing in environment variables.
Allow users to set environment variables, as desired in https://github.com/bencherdev/bencher/issues/416.

For example:

```sh
bencher run --env BENCHER_CONFIG_PATH=./alt/path/to/config.json
```

It should have the same semantics as `-e` and `--env` for `docker run`: https://docs.docker.com/reference/cli/docker/container/run/#env

epompeii:
Completed: https://bencher.dev/docs/reference/changelog/#v0413

Issue #416, opened 2024-06-12
BENCHER_CONFIG_PATH env variable for docker self-hosted instance


ricardaxel:
Hello, 

I believe that `BENCHER_CONFIG_PATH` / `BENCHER_CONFIG` env variables are not passed to `bencher_api` container when using `bencher up`.

For example, this won't use my_config.json file 

```
>> export BENCHER_CONFIG_PATH=my_config.json
>> bencher up
```

I think this is because [the config to create a container](https://github.com/bencherdev/bencher/blob/main/services/cli/src/bencher/sub/docker/up.rs#L140-L147) doesn't have the [env](https://github.com/fussybeaver/bollard/blob/master/src/container.rs#L176) entry set. Am I missing something ?

epompeii:
Hey @ricardaxel ,
Thanks for checking out Bencher. I'm sorry you ran into trouble.

> I believe that `BENCHER_CONFIG_PATH` / `BENCHER_CONFIG` env variables are not passed to `bencher_api` container when using `bencher up`.

You are correct. They are not passed into the container.

> I think this is because [the config to create a container](https://github.com/bencherdev/bencher/blob/main/services/cli/src/bencher/sub/docker/up.rs#L140-L147) doesn't have the [env](https://github.com/fussybeaver/bollard/blob/master/src/container.rs#L176) entry set. Am I missing something ?

I think you are spot on! I have created a tracking issue to add this functionality: https://github.com/bencherdev/bencher/issues/417
I'm open to PRs if you would like to take a crack at it 😃 Otherwise, will try to get to it in the near future.

In the mean time, you can use this [docker-compose.yml file](https://github.com/bencherdev/bencher/blob/main/docker-compose.yml) as a starting point for a more advanced setup.

ricardaxel:
Thanks for the response ! Sure I can do a PR

epompeii:
@ricardaxel thank you for the PR!

I ended up making a few tweaks to things so that you can pass environment variables and set volumes for the containers individually. The details are in [the changelog](https://bencher.dev/docs/reference/changelog/#v0413).

PR #418, opened 2024-06-13
Bencher up cli : add --env and --volume options

ricardaxel:
Solve #416 #417

I also added --volume, so that we can pass config file to the container. The idiomatic usages would be : 

```
>> bencher up --env BENCHER_CONFIG="$(cat config.json)"
>> bencher up --volume "$(pwd):/bencher" --env BENCHER_CONFIG_PATH=/bencher/config.json
```


epompeii:
Great! Thank you, @ricardaxel.

This will be included in the upcoming release.

PR #532, opened 2024-11-11
fix: Write summary into the filepath provided by GITHUB_STEP_SUMMARY 

### LINKED ISSUE #530: bug: GitHub Summary implementation does not work
The latest release 0.4.25 ‘s GitHub Summary does not work atm. Pointed out by @sagudev this is due to the `$GITHUB_STEP_SUMMARY` is actually a path to a file, so the command `echo “Summary” >> $GITHUB_STEP_SUMMARY` actually writes into the appointed file instead of that environment variable.

I can put out a tested PR to fix this as well.

dklassic:
fixes #530

A quick fix to make GitHub Summary work. [I've tested it here](https://github.com/dklassic/bencher-test/actions/runs/11774831460) showing that it is working now.


epompeii:
Great, thank you @dklassic!

PR #567, opened 2025-02-07
chore: add JetBrains and arch support to devcontainers

brianheineman:
Adds support for JetBrains IDEs (tested w/RustRover) and `arm64` architectures through the `ARCH` environment variable.

image


epompeii:
Great, thank you @brianheineman!

Issue #564, opened 2025-02-06
Add logarithmic scaling to graphs


brianheineman:
Some of the graphs become difficult to reason on when there are significant differences in benchmark values.  Please consider adding logarithmic scaling to the graphs to make them easier to reason on.

Example of a graph that would benefit from this feature: https://github.com/theseus-rs/file-type/issues/71#issuecomment-2638317887

epompeii:
Completed: https://bencher.dev/docs/reference/changelog/#v0435

kaimast:
Is there a way to manually set graphs to log scale, or change the ratio at which it switches? I have a graph with about 9x difference, and log scale would be very helpful here.

epompeii:
> Is there a way to manually set graphs to log scale, or change the ratio at which it switches?

@kaimast currently there is not. I've created a tracking issue for this though: https://github.com/bencherdev/bencher/issues/917

PR #570, opened 2025-02-07
feat: charts use pow scaling to allow users to more easily reason on

brianheineman:
Update charts use pow scaling to allow users to more easily reason on graphs with highly differentiated values.  Previously, charts were displaying values using a linear scale; e.g.:

image

**Before: linear**
image

**After: pow (.5) if max > 1000x min**
image

**After: pow (1) if min < 1000x**
image

**After: pow (1) if min < 1000x (large value not selected)**
image

Addresses: #564

### DIFF
--- a/services/console/src/components/console/perf/plot/line/LinePlot.tsx
+++ b/services/console/src/components/console/perf/plot/line/LinePlot.tsx
@@ -439,7 +439,7 @@ type Scale = {
 	measure: JsonMeasure;
 	factor: number;
 	units: string;
-	yScale: d3.ScaleLinear<number, number, never>;
+	yScale: d3.ScalePower<number, number, never>;
 };
 
 const scale_data = (
@@ -491,7 +491,9 @@ const get_scale = (
 	const units = scale_units(min, raw_units);
 
 	const domain = [min / factor, max / factor];
-	const yScale = d3.scaleLinear().domain(domain).nice();
+	// Use pow scaling to allow users to more easily reason on graphs with highly differentiated values
+	// See: https://observablehq.com/plot/features/scales#continuous-scales
+	const yScale = d3.scalePow().exponent(0.5).domain(domain).nice();
 
 	return {
 		measure,
--- a/services/console/src/components/console/perf/plot/line/LinePlot.tsx
+++ b/services/console/src/components/console/perf/plot/line/LinePlot.tsx
@@ -490,10 +490,15 @@ const get_scale = (
 	const factor = scale_factor(min, raw_units);
 	const units = scale_units(min, raw_units);
 
-	const domain = [min / factor, max / factor];
-	// Use pow scaling to allow users to more easily reason on graphs with highly differentiated values
+	// Use pow scaling to allow users to more easily reason on graphs with
+	// highly differentiated values. If the min is less than 1,000 times
+	// smaller than max use a linear scale pow(1), or else use a square root
+	// scale pow(.5).
+	//
 	// See: https://observablehq.com/plot/features/scales#continuous-scales
-	const yScale = d3.scalePow().exponent(0.5).domain(domain).nice();
+	const exponent = (max / min) < 1000 ? 1 : 0.5;
+	const domain = [min / factor, max / factor];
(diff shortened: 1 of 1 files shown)


brianheineman:
With extreme values, trying to set the exponent to a low of a value (.1) causes labels to overlap; a better configuration is likely required for handling extreme min/max ranges:

image

services/console/src/components/console/perf/plot/line/LinePlot.tsx
```diff
@@ -490,8 +490,15 @@ const get_scale = (
 	const factor = scale_factor(min, raw_units);
 	const units = scale_units(min, raw_units);
 
+	// Use pow scaling to allow users to more easily reason on graphs with
+	// highly differentiated values. If the min is less than 1,000 times
+	// smaller than max use a linear scale pow(1), or else use a square root
+	// scale pow(.5).
+	//
+	// See: https://observablehq.com/plot/features/scales#continuous-scales
+	const exponent = (max / min) < 1000 ? 1 : 0.5;
```
epompeii, line 499: Great start! I think we're going to need to get a bit more clever with how we calculate the `exponent`. There is probably a nifty algorithm for this someone has come up with.

epompeii: Testing things out with a little bit of 🤖 help, this seems to work pretty well:

```typescript
const relativeDifference = max / min;
const exponent = Math.max(1 / Math.log10(relativeDifference), 0.3333);
```

epompeii: `bencher mock --pow 15`

Screenshot 2025-02-07 at 11 29 08 PM

`bencher mock --pow 18`

Screenshot 2025-02-07 at 11 31 05 PM

`bencher mock --pow 30`

Screenshot 2025-02-07 at 11 33 33 PM

epompeii: Okay, playing around more and this is the updated version:

```typescript
const relativeDifference = max / min;
const exponent = Math.max(
  relativeDifference === 1
    ? relativeDifference
    : 1 / Math.log10(relativeDifference),
  0.3333,
);
```

If you only have one value, the `max` and `min` are the same so `relativeDifference` is `1`, and `log10(1)` is `0`, which leads to `1 / 0`. No good!

brianheineman: image

The docs provide that as one of the examples; it creates a "wider" band at the lower end.  Based on some of the testing I did, I think you could even go slightly lower (.25) and it looks reasonable; too much lower than that and the axis labels start rendering poorly.  Let me know if you want me to update it to `1 / 3`.  I think values between `.3` and `.5` are a significant improvement.

epompeii: Ah, good to know! So I slept on this and then played around a bit more.

This is what I've come up with:

```typescript
const relativeDifference = max / min;
const exponent =
  relativeDifference < 10
    ? 1
    : Math.max(1 / Math.log10(relativeDifference), 1 / 3);
```

`relativeDifference < 10`

If the relative difference is less than 10x, then we just use the linear scale. This also solves the divide by zero condition and keeps the graph from contracting at the bottom.

`1 / Math.log10(relativeDifference)`

The `Math.log10(10)` is `1` so this really gives us the entire scale where IMO it makes sense. That is, we wont ever contract the bottom of the graph.

`Math.max(..., 1 / 3)`

This makes sure we don't go too far and get any overlapping y-axis ticks. I checked `.25`, and it leads to slight overlap/touching on some of my test plots:

`1 / 4`

Screenshot 2025-02-08 at 7 59 06 AM

`1 / 3`

Screenshot 2025-02-08 at 7 59 22 AM

What do you think?

brianheineman: I like it; I made the change and it is working well with the tests I was running also.

image
image
image

From what I can tell it will go to square root @ ~100x different and down to no more than 1/3; this is the range I was seeing work well.

epompeii:
Great! Thank you, @brianheineman!

Issue #561, opened 2025-02-05
Add `cargo-deny`


epompeii:
Add [cargo-deny](https://github.com/EmbarkStudios/cargo-deny) to workspace and use [the GitHub Action](https://github.com/EmbarkStudios/cargo-deny-action).

PR #571, opened 2025-02-07
build: add cargo deny check

brianheineman:
Addresses #561

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,53 @@
+# Documentation for this configuration file can be found here
+# https://embarkstudios.github.io/cargo-deny/checks/cfg.html
+
+[graph]
+targets = [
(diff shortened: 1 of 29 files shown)


deny.toml
```diff
@@ -0,0 +1,53 @@
...
+
+[advisories]
+ignore = [
+    "RUSTSEC-2024-0370",
+    "RUSTSEC-2024-0384",
+    "RUSTSEC-2024-0399",
+    "RUSTSEC-2024-0402",
+    "RUSTSEC-2024-0421",
+    "RUSTSEC-2025-0004",
+]
```
brianheineman, line 23: Ignoring the advisories for now so that cargo deny passes.  However, they should be reviewed, and likely there are several dependencies that should be updated.

epompeii:
Great, thank you @brianheineman!

PR #573, opened 2025-02-08
fix: address RUSTSEC-2024-0370 by updating tabled to 0.18

brianheineman:


epompeii:
Thank you, @brianheineman!

Issue #572, opened 2025-02-08
Check Typeshare generation in CI


epompeii:
All of the other generated types are checked in CI. At some point, I stopped checking [Typeshare](https://github.com/1Password/typeshare) as we are not yet running Typescript type checking during builds: https://github.com/bencherdev/bencher/issues/557

However, having different versions of Typeshare leads to different outputs which can be confusing for new contributors.

Unify all of the Typeshare versions that appear in the code base:

1. https://github.com/bencherdev/bencher/blob/1e9ab250219a344e9e94c5e1ea8d20cead41c11f/.devcontainer/Dockerfile#L50
2. https://github.com/bencherdev/bencher/blob/1e9ab250219a344e9e94c5e1ea8d20cead41c11f/services/console/src/types/bencher.ts#L2

It would probably best to just go ahead and update to the latest version while doing so: https://github.com/1Password/typeshare/releases/tag/v1.13.2

Then update the CI check: https://github.com/bencherdev/bencher/blob/1e9ab250219a344e9e94c5e1ea8d20cead41c11f/.github/workflows/bencher.yml#L177-L196

It will need to install the `typeshare-cli` in the same manner as the dev container:

```shell
cargo install typeshare-cli --version 1.13.2 --locked
```

Then generate the types, with the workspace alias:

```shell
cargo gen-ts
```

Finally, run the git check:

```shell
git diff --exit-code
```

PR #575, opened 2025-02-08
build: check typescript type generation

brianheineman:
Addresses #572


.github/workflows/bencher.yml
```diff
@@ -186,6 +186,12 @@ jobs:
         run: cargo gen-spec
       - name: Check OpenAPI Spec for changes
         run: git diff --exit-code
+      - name: Generate Typescript Types
+        run: |
+          cargo install typeshare-cli --version 1.13.2 --locked
```
epompeii, line 191: Let's break this out into a separate step. It makes it a lot easier to glance and see what is going on in GitHub Actions.

epompeii:
Great! Thank you, @brianheineman!

PR #617, opened 2025-08-02
fix permission on track gh action example

ruyadorno:


epompeii (approved):
Thank you, @ruyadorno !

Issue #619, opened 2025-09-22
Alias for `rust_iai_callgrind`


gamma0987:
Hi. I've just renamed the iai-callgrind project to gungraun. Thanks for your suggestion to do so by the way. You mentioned you can add an alias for `rust_iai_callgrind`? The terminal output format is basically the same besides a small change in the summary line from `Iai-Callgrind result: ...` to `Gungraun result: ...`.

epompeii:
> Can you send me an invite link to the server?

You can get an invite by clicking here: https://bencher.dev/chat

> If there's a bug in the parser, I can take a look at it and fix it in a pr.

Yes, I believe it is a bug in the parser where it is failing to collect some of the benchmarks. And a fix PR would be great!

gamma0987:
Sure :) I added a pr for the test. Thanks for the alias.

epompeii:
Completed in `v0.5.6`: https://github.com/bencherdev/bencher/releases/tag/v0.5.6

Updated docs: https://bencher.dev/docs/explanation/adapters/#-rust-gungraun

PR #620, opened 2025-09-25
fix: benchmark names with multiple lines

gamma0987:
Here's the fix for the problem on discord https://discord.com/channels/1074890101362004088/1412794833072033903/1413138826645209198 and as discussed in #619. Because there's no github issue: Here's a quick summary of the problem and the solution:

In the output of the slang benchmark there are benchmarks which have a multiline benchmark name. These benchmarks are not parsed correctly and are skipped by the current parser which causes the discrepancy between the expected amount of `12` benchmarks and the actual amount of `7` in this benchmark run https://github.com/NomicFoundation/slang/actions/runs/17408413743/job/49418989630. This issue arises in gungraun due to the use of syn's `to_string` function, which occasionally splits strings across multiple lines. While this behavior may not be considered a bug within gungraun, it deviates from the original intention of keeping both the benchmark name and description on a single line.

This pr contains a fix for the parser to accept benchmark names on multiple lines. Multiline benchmark names and descriptions will be removed in future versions of gungraun but this fix should work for all versions of gungraun/iai-callgrind.

### DIFF
--- a/lib/bencher_adapter/src/adapters/rust/iai_callgrind.rs
+++ b/lib/bencher_adapter/src/adapters/rust/iai_callgrind.rs
@@ -10,9 +10,9 @@ use bencher_json::{
 use nom::{
     IResult,
     branch::alt,
-    bytes::complete::{is_a, is_not, tag},
+    bytes::complete::{is_a, is_not, tag, take_until1},
     character::complete::{char, space0, space1},
-    combinator::{map, opt, recognize},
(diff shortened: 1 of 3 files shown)


epompeii (approved):
@gamma0987 thank you for PR!
This all looks good to me. The CI failures are expected. I need to update the workflows to be more fork PR friendly.

PR #624, opened 2025-11-11
fix: Allow null for a dotnet namespace and add a small test

### LINKED ISSUE #605: c_sharp_dot_net adapter fails if namespace is null
I didn't want the namespace to be in the name of my benchmark results in bencher, so i removed it from my code. However, it seems the adapter must have it. I think it would be prudent to support no namespace.

To replicate, just set namespace to null in the [unit test file](https://github.com/bencherdev/bencher/blob/main/lib/bencher_adapter/tool_output/c_sharp/dot_net/two.json).

sukinathan:
Fixes #605. Let me know if any changes are needed.


epompeii (changes requested):
@ntransw thank you for taking the time to make this fix and to add a test!

Just one small tweak, other than that looks good.

epompeii:
Thank you, @ntransw!
I'll close out https://github.com/bencherdev/bencher/issues/605 once these changes are included in the next release.

PR #671, opened 2026-02-16
Nix flake

MathisWellmann:
[Nix] is a declarative approach to creating reproducible packages and artifacts and makes it easy for users to run the `bencher` software.

## Changes include:
- Introduce the nix package for `bencher` which can be run directly from the github repo with `nix run github:bencherdev/bencher` or simply locally with `nix run`.
- Also include nix package for `api`, which can also be run with `nix run .#api`.
- The flake also includes a reproducible development shell which can be activated with `nix develop` or automatically with `direnv allow`, if `direnv` is installed.
- CI: add `build_bencher_nix` step.
- CI: add `build_api_nix` step.
- CI: add `nix_lints` step, which runs `alejandra`, `statix` and `deadnix` to ensure clean nix code.

[Nix]: https://nixos.org/

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -108,3 +108,19 @@ jobs:
         run: |
           npm install --include=dev
           npx biome ci --formatter-enabled false --organize-imports-enabled false .
+
+  nix_lints:
+    name: Check nix code
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v6
+      - name: Install Nix
+        uses: cachix/install-nix-action@v25
+        with:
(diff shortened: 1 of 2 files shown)


epompeii (commented):
Hey @MathisWellmann thank you for the PR to add Nix support to Bencher.

Would it make sense to add a job to [the `build.yml` workflow](https://github.com/bencherdev/bencher/blob/main/.github/workflows/build.yml) to make sure things actually build in CI?

Further, we probably want to add a task under `tasks` for updating the `flake.lock` and check that in CI too.

epompeii:
Thank you, @MathisWellmann!

PR #742, opened 2026-03-27
Add benchmark aliases for multiple exact-match identifiers

### LINKED ISSUE #674: Add multiple aliases for a Benchmark
A Benchmark can now be identified by its name, slug, or UUID: https://github.com/bencherdev/bencher/issues/629

This helps align Benchmarks with the other dimensions (Branch, Testbed, and Measure).

It would be great to be able to track multiple aliases for a Benchmark over time with an array of exact match strings.

```json
[
  "current_benchmark::name",
  "old_benchmark::name",
]
```

The `name` field could then become a "display name" similar to https://github.com/bencherdev/bencher/issues/366.
The current name would just be added as the first entry in the match array.

Louisvranderick:
## Summary
- Add a new `benchmark_alias` table with `UNIQUE(project_id, alias)` and cascading foreign keys so benchmarks can keep multiple exact-match aliases over time.
- Resolve benchmark `NameId::Name` by primary name or alias (single query with `OR EXISTS`), and validate alias uniqueness inside create/update write transactions to avoid TOCTOU races.
- Expose aliases through API + generated types, wire CLI create/update `--alias`, and batch-load aliases in list/report paths to avoid N+1 query behavior.

Closes #674

## Test plan
- [x] `cargo fmt`
- [x] `cargo clippy -p bencher_schema -p api_projects -p bencher_cli --no-deps --all-targets --all-features -- -Dwarnings`
- [x] `cargo test -p bencher_schema model::project::benchmark`
- [x] `cargo check --no-default-features`
- [x] `cargo gen-types` outputs committed (`services/api/openapi.json`, `services/console/src/types/bencher.ts`)

Made with [Cursor](https://cursor.com)


lib/bencher_json/src/project/benchmark.rs
```diff
@@ -48,6 +51,9 @@ pub struct JsonBenchmark {
     pub created: DateTime,
     pub modified: DateTime,
     pub archived: Option<DateTime>,
+    /// Additional exact-match strings (the primary [`Self::name`] is always a matcher).
+    #[serde(default)]
+    pub aliases: Vec<BenchmarkName>,
```
epompeii: Conceptually, `aliases` should come after `slug` and before the timestamp fields.

lib/bencher_schema/src/model/project/benchmark/mod.rs
```diff
@@ -50,13 +62,39 @@ impl QueryBenchmark {
...
+        conn: &mut DbConnection,
+        project_id: ProjectId,
+        name_id: &BenchmarkNameId,
+    ) -> Result<Self, HttpError> {
+        match name_id {
+            NameId::Uuid(_) | NameId::Slug(_) => schema::benchmark::table
+                .filter(schema::benchmark::project_id.eq(project_id))
+                .filter(Self::eq_name_id(name_id))
+                .first::<Self>(conn)
+                .map_err(resource_not_found_err!(Benchmark, (project_id, name_id))),
```
epompeii: If we're already matching here, it is a bit silly to then have to requery by `NameId`.

epompeii (changes requested):
Thank you @Louisvranderick! Just took a cursory look.

epompeii:
@Louisvranderick sorry I just realized you had replied and pushed up changes!

Please, for future iterations re-request review, so I get pinged that I need to take another look:
Screenshot 2026-05-23 at 03 55 26

I'll try to get to this soon, now that it is back on my radar!

services/cli/src/parser/project/benchmark.rs
```diff
@@ -66,6 +66,10 @@ pub struct CliBenchmarkCreate {
     #[clap(long)]
     pub slug: Option<BenchmarkSlug>,
 
+    /// Additional exact-match alias (repeat for multiple)
+    #[clap(long = "alias")]
+    pub aliases: Vec<BenchmarkName>,
```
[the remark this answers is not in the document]
epompeii: @Louisvranderick is there a reason you did not want to make this change?

PR #923, opened 2026-07-07
Decrease run-to-run variance of the bare metal runner

epompeii:
## What

Reduces run-to-run variance of the bare metal `runner` on both the Firecracker sandbox path and the no-sandbox path. All new measures are runtime sysfs/procfs/cgroup configuration shipped inside the single static `runner` binary (no host boot args required), on by default with individual disable flags, best-effort with graceful skips on ARM and locked-down hosts, and restored on shutdown.

One commit per phase:

### 1. Host tuning extensions, `runner run` fix, preflight warnings

- New tuning knobs (each with a keep-enabled CLI flag): NUMA balancing, timer migration, soft watchdog, and KSM disabled via sysctls; deep C-states held off via a `/dev/cpu_dma_latency` PM QoS fd held open for the runner lifetime (crash-safe, works on x86 and ARM)
- Device IRQs, the IRQ default affinity, and the unbound workqueue cpumask steered to housekeeping cores (`--no-irq-steering` to disable), applied via a new layout-scoped tuning phase after `CpuLayout::detect()`
- Bug fix: one-shot `runner run` applied tuning but never detected the CPU layout, so cpuset pinning only happened under `runner up`; it now mirrors the `up` path
- Static preflight warnings at startup: irqbalance running, no cgroup v2, virtualized host, missing `isolcpus=`/`nohz_full=` boot args

### 2. No-sandbox path isolation and metrics

- The local execution path previously ran the benchmark with no CPU isolation at all; it now gets a per-run cgroup cpuset on the benchmark cores, with race-free attachment via `Command::pre_exec` (the child moves itself into the pre-opened `cgroup.procs` fd between fork and exec) and a `sched_setaffinity` fallback for unprivileged hosts
- The local path now emits `RunMetrics` (wall clock plus cgroup CPU/memory, `transport: "local"`); it previously emitted none
- On timeout or cancellation, surviving grandchildren are reaped via `cgroup.kill` so no stray work lingers on benchmark cores and the per-run cgroup can be removed

### 3. Scheduler-domain isolation and vCPU determinism (Firecracker)

- `/sys/fs/cgroup/bencher` becomes an `isolated` cpuset partition, the runtime equivalent of `isolcpus=`: benchmark cores leave the root scheduling domain so nothing else can be balanced onto them. Kernel acceptance is verified by read-back, with a fallback chain `isolated` -> `root` -> `member` (`--no-cpuset-partition` to disable). Cpuset mems mirror the parent's effective memory nodes instead of hardcoding node 0, keeping multi-node NUMA hosts on their local nodes
- Each Firecracker `fc_vcpu N` thread is pinned to its own dedicated benchmark core after `InstanceStart` (required companion to the isolated partition, which does no load balancing); VMM/API threads pin to the last benchmark core
- Swap is disabled for the VM cgroup in the live path, and `RunMetrics.cgroup` is now filled from the existing cgroup metrics reader (previously always `None`)

### 4. Guest determinism and host THP

- Guest kernel cmdline gains `norandmaps nokaslr transparent_hugepage=never` (guest ASLR/KASLR off, deterministic guest memory backing), and is now arch-conditional: aarch64 uses `keep_bootcon` and `reboot=k` instead of the x86-only `reboot=t`
- Host transparent hugepages default to `never` (`enabled` and `defrag`), with `--thp <never|madvise|always|leave>`; `leave` preserves the old do-not-touch behavior

> [!NOTE]
> Release notes for operators:
> - All new measures are on by default. Existing `runner up` deployments will start steering IRQs, disabling NUMA balancing/watchdogs/KSM, holding C-states off, creating an isolated cpuset partition, and pinning host THP to `never` on upgrade. Each has an individual disable flag, and `--no-tuning` disables everything.
> - The host THP default change may cause a one-time baseline step for memory-heavy benchmarks. `--thp leave` opts out.
> - Tuning restores on clean shutdown only. After a SIGKILL or panic-abort, sysctls/IRQ affinities/THP/partition stay applied and the pre-tuning values cannot be recovered by the next run; a reboot or manual reset recovers them. The C-state hold and per-run cgroups self-heal.

## Testing

- Unit tests for every new helper (tempfile-based sysfs/procfs trees), 208 tests passing in `bencher_runner` + `bencher_runner_cli`
- `cargo clippy -Dwarnings` natively and cross-compiled to `x86_64-unknown-linux-gnu` (`./scripts/clippy.sh`), `cargo check --no-default-features`, `cargo test --doc`, `./scripts/test.sh --linux-only`
- Operator-side validation still to come on Linux/KVM hardware: `cargo test-runner`, plus before/after `bencher noise` and repeated `runner run --iter N` on x86_64 and aarch64, with and without `--sandbox firecracker`

### DIFF
--- a/plus/bencher_runner/src/config.rs
+++ b/plus/bencher_runner/src/config.rs
@@ -175,8 +175,26 @@ fn default_disk() -> Disk {
     DEFAULT_DISK
 }
 
+/// Default guest kernel command line.
+///
+/// Shared determinism params on both architectures: `norandmaps` disables
+/// guest ASLR (opt-out parity with the host `--aslr` flag via the
+/// user-overridable `kernel_cmdline`), `nokaslr` disables guest KASLR
+/// (every run boots a fresh VM, so KASLR is per-run variance unique to
+/// the sandbox path), and `transparent_hugepage=never` makes guest memory
+/// backing deterministic.
+///
+/// Arch differences: `reboot=t` (triple fault) is x86-only; aarch64 uses
+/// `reboot=k` and needs `keep_bootcon` for console output during boot.
 fn default_kernel_cmdline() -> String {
-    "console=ttyS0 reboot=t panic=1 pci=off root=/dev/vda rw init=/init".to_owned()
+    #[cfg(target_arch = "aarch64")]
+    {
+        "keep_bootcon console=ttyS0 reboot=k panic=1 pci=off root=/dev/vda rw init=/init norandmaps nokaslr transparent_hugepage=never".to_owned()
+    }
+    #[cfg(not(target_arch = "aarch64"))]
+    {
+        "console=ttyS0 reboot=t panic=1 pci=off root=/dev/vda rw init=/init norandmaps nokaslr transparent_hugepage=never".to_owned()
+    }
 }
 
 const fn default_timeout_secs() -> u64 {
@@ -557,5 +575,37 @@ mod tests {
     #[test]
     fn config_with_empty_token_errors() {
         Config::new("img").with_token("").unwrap_err();
+    }
+
+    #[test]
+    fn default_kernel_cmdline_guest_determinism() {
+        let config = Config::new("img");
+        assert!(config.kernel_cmdline.contains("norandmaps"));
+        assert!(config.kernel_cmdline.contains("nokaslr"));
+        assert!(config.kernel_cmdline.contains("transparent_hugepage=never"));
+        assert!(config.kernel_cmdline.contains("init=/init"));
+    }
+
+    #[cfg(target_arch = "aarch64")]
+    #[test]
+    fn default_kernel_cmdline_aarch64() {
+        let config = Config::new("img");
+        assert!(config.kernel_cmdline.starts_with("keep_bootcon "));
+        assert!(config.kernel_cmdline.contains("reboot=k"));
+        assert!(!config.kernel_cmdline.contains("reboot=t"));
+    }
+
+    #[cfg(not(target_arch = "aarch64"))]
+    #[test]
+    fn default_kernel_cmdline_x86() {
+        let config = Config::new("img");
+        assert!(config.kernel_cmdline.contains("reboot=t"));
+        assert!(!config.kernel_cmdline.contains("keep_bootcon"));
+    }
+
+    #[test]
+    fn kernel_cmdline_override_preserved() {
+        let config = Config::new("img").with_kernel_cmdline("console=ttyS0 custom=1");
+        assert_eq!(config.kernel_cmdline, "console=ttyS0 custom=1");
     }
 }
--- a/plus/bencher_runner/src/cpu.rs
+++ b/plus/bencher_runner/src/cpu.rs
@@ -222,12 +222,13 @@ fn format_cpuset(cpus: &[usize]) -> String {
 /// `/sys/devices/virtual/workqueue/cpumask`. Words are ordered most
 /// significant first; all words after the first are zero-padded to
 /// 8 hex digits (e.g., `[0, 1]` -> `"3"`, `[32]` -> `"1,00000000"`).
+#[cfg(target_os = "linux")]
 #[must_use]
 #[expect(
     clippy::integer_division,
     reason = "dividing CPU IDs into 32-bit mask words"
 )]
-pub fn format_cpumask(cpus: &[usize]) -> String {
+pub(crate) fn format_cpumask(cpus: &[usize]) -> String {
     use std::fmt::Write as _;
 
     let Some(&max) = cpus.iter().max() else {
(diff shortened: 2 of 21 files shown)


epompeii:
Addressed the review asks in be106d0 and 3d12eea:

- Fixed the partition restore gap: the restore entry is now saved before any mode write, normalized to the mode token, with a regression test for the dirty-file case (be106d0)
- Made enable_controllers verification-gated so pre-configured unprivileged runners keep working, and fixed the controller check to match whole tokens (be106d0)
- Documented the crash-restore limitation in the tuning module docs, the zero-only /dev/cpu_dma_latency encoding, and the cpuset-vs-pinning guarantee (3d12eea)
- Expanded the PR description release notes to cover the on-by-default behavior change for existing runners

Confirming the compliance asks: cargo fmt, cargo clippy -Dwarnings (native and cross-compiled via ./scripts/clippy.sh), cargo nextest run (193 passing), cargo test --doc, cargo check --no-default-features, and ./scripts/test.sh --linux-only are all green locally on every commit.

Not changed: guest ASLR stays cmdline-only (norandmaps via the user-overridable kernel_cmdline was a deliberate design decision), and --thp keeps FromStr parsing to match the existing SandboxLogLevel convention.

epompeii:
Round 3 review nits addressed in 3f4020e (comment-only):

- CliTuning now documents the flag polarity rationale: keep-enabled flags name hardware features (--smt, --turbo, --cstates), --no- flags name skipped runner actions (--no-irq-steering, --no-cpuset-partition), since a bare --irq-steering would read as enabling something already on by default
- assign_core documents that at full vCPU width the highest-index vCPU shares its core with VMM threads
- runner run documents that core pinning is deliberately independent of --no-tuning, matching the long-standing runner up behavior (the tuning config gates the tuning knobs, including partition and steering); confirming this is intended
- the local spawn-failure path documents why it emits no metrics (nothing ran)

On the workqueue scope of --no-irq-steering: the flag's own help text already reads "Do not steer device IRQs and kernel workqueues to housekeeping cores", so the scope is documented at the point of use; keeping the shorter flag name.

Linux cross-checks (./scripts/clippy.sh, ./scripts/test.sh --linux-only) plus nextest, doc tests, and no-default-features check remain green on this commit; both the unqualified size_of in local_isolation.rs and the unconditional CpuLayout import compile on both targets.

epompeii:
Round 4 review points, addressed in e07f278:

- Point 1 (concurrent runners): added the suggested doc note; tuning assumes a single runner process per host since the mutated state is host-global
- Point 2 (--no-tuning scope): confirmed intended. --no-tuning disables host-global mutations; per-run core pinning and the transient per-run cgroup are core runner behavior, matching what runner up has always done. Documented in run.rs
- Point 4 (transport "local"): confirmed safe. The only consumer that inspects the transport value is the sandboxed-only metrics_transport_type scenario, and RunMetrics deserializes transport as a plain string. Also added a new nosandbox_metrics scenario to cargo test-runner that pins down the local-path marker and transport value
- Points 5 and 6: leaving as is. The 'pinned 0 of N' summary is deliberate signal that pinning failed, and FromStr parsing for --thp matches the existing SandboxLogLevel convention

Cross-compile checks, nextest, doc tests, and no-default-features remain green on this commit.

epompeii:
Round 5 observations, addressed in 8c233dd:

- Observation 2 (swap asymmetry): fixed. LocalIsolation now also disables swap on the per-run cgroup, so the local path fully mirrors the Firecracker path
- Observation 1 (enable_controllers on pre-existing parents): confirmed intended. The verification read is the gate, so unprivileged runners with admin-pre-enabled controllers still succeed; a parent that exists but lacks required controllers would previously have failed later at the first cgroup write anyway, just with a worse error
- Observations 3 and 4: known, documented constraints (last-core sharing at full vCPU width; single-runner-per-host assumption)

Cross-compile checks (./scripts/clippy.sh, ./scripts/test.sh --linux-only), nextest, doc tests, and no-default-features are green on this commit.

epompeii:
Round 6 findings, addressed in 98696e5:

- Finding 1 (emdash in the relocated comment): fixed, along with the same pre-existing comment in up/mod.rs
- Finding 3 (mixed separators in tuning output): normalized all output strings in the tuning module to hyphens per the CLAUDE.md no-emdash policy
- Finding 2 (contiguous core ID assumption): fixed now rather than deferred, since this PR's SMT-off default is what creates the non-contiguous condition. CpuLayout::detect() parses the actual ID list from /sys/devices/system/cpu/online (with the count-based path kept as fallback), so interleaved-sibling topologies pin to real online cores. New with_cpu_ids constructor with tests for the interleaved case
- Runner docs note on --no-tuning pinning: leaving for the docs site separately; the PR description and code comments cover it

epompeii:
Round 7 review: all observations are known, documented constraints or no-action items. Two follow-ups:

- The concurrent-runner mutual-clobbering hardening is now tracked as #924 (advisory flock before tuning applies), per the suggestion to make it a follow-up item rather than a docs-only note
- Confirming the CI question: the Test / Cargo Test job runs the workspace test suite and the Runner jobs build both Linux targets; the cross-compile lint scripts (./scripts/clippy.sh, ./scripts/test.sh --linux-only) are additionally run locally on every commit of this branch, and the non-Linux expect attributes compile clean on both targets

CI is green on this branch apart from the two pre-existing online lint failures (Lint / Zizmor Online, Lint / Cargo Deny Advisories), which also fail on devel.

PR #927, opened 2026-07-08
Add runner update channels for canary field testing on Bencher Cloud

epompeii:
## What

Adds runner update channels so `runner` changes can be field-tested on Bencher Cloud before a versioned release, mirroring how the API server is dogfooded via `cloud` branch deploys.

- New `runner up --update-channel <stable|canary>` flag (env `BENCHER_UPDATE_CHANNEL`, default `stable`, conflicts with `--no-auto-update`)
- The `stable` channel keeps today's behavior: exact version match against the server, updates from versioned GitHub Releases
- The `canary` channel converges by content: the runner reports the SHA-256 of its own binary in `Ready` metadata, and the server compares it against the published checksum of a rolling `canary` prerelease (TTL-cached per architecture), sending the existing `Update` directive on mismatch

## How

- **Protocol**: `JsonRunnerMetadata` gains optional `channel` and `checksum` fields; absent means `stable`, so old runners and old servers interoperate unchanged and the stable wire format is byte-identical
- **Server**: new `RunnerUpdate` context (URL construction + canary checksum cache with injectable `Clock`), channel-aware `handle_ready` with an injectable checksum fetcher for unit testing, and a new `runners.update_base_url` config knob. Checksum fetch failures now log, count (`runner.self_update.check.failed`; updates count as `runner.self_update.sent`), and skip the update offer instead of erroring the channel into a reconnect loop (this also fixes the stable path when GitHub is unreachable)
- **CI**: the runner build now runs on `cloud` pushes (path-filtered, `build-only` skips the KVM scenario tests already run on devel) and is versioned `canary`; a new `publish_runner_canary` job clobbers the assets on a rolling `canary` prerelease, binaries before checksums so a partial upload never advertises a missing binary. The tag is `canary` rather than `cloud` because a tag named like the branch is ambiguous in git ref resolution and would break `git checkout cloud` in `rebase_cloud.yml`
- **Ops**: `cargo ops deploy|start --update-channel` and an optional `update_channel` field in `runners.json` write `BENCHER_UPDATE_CHANNEL` into the systemd drop-in
- **Docs**: `runner up` reference, Runner Protocol reference, and the Self-Hosted Runners explanation updated in all 9 languages; changelog entries added under `Pending v0.6.9`

Self-update remains idle-only on both sides (server sends `Update` only from `wait_for_ready`; the runner handles it only in `AwaitingJob`), so a canary rollout never interrupts a running benchmark.

## Field testing flow

1. Merge a runner change to `devel`, reset `cloud` to `devel`, and push
2. Cloud CI publishes the new canary assets while the API deploys as usual
3. At their next idle poll, canary channel runners detect the checksum mismatch, self-update, and reconnect
4. Rollback is re-uploading previous assets with `--clobber` or pushing a revert to `cloud`


epompeii:
Review follow-up on the five observations:

1. Confirmed correct: the changes filter has no explicit base, so cloud pushes diff against the default branch (`main`), which CI resets to `cloud` after each successful deploy. That yields "changed since the last deploy" semantics. Now documented in a comment on the `runner` job.
2. Intentional asymmetry; added a note to the `stable_update` doc comment.
3. No action needed, as noted.
4. Keeping the direct `UpdateChannelKind` constructions: inside the per-channel functions they are clearer than routing through the shared helper.
5. Keeping the tuple in `runner_ops`: it is an internal helper in an admin task crate and the types are distinct.

epompeii:
Follow-up on the latest review round:

1. Fixed: `RunnerSelfUpdateSent` now increments after a successful `tx.send`, so a serialization or send failure no longer counts as a sent update.
2. Acknowledged as designed: a canary runner without a self-checksum stays put, with warnings on both sides.
3. As documented: the stable path intentionally skips caching.
4. Keeping the wasm export for consistency with sibling validators.

Local gates confirmed: `cargo clippy --all-features -- -Dwarnings` and `cargo check --no-default-features` are both clean.

epompeii:
On observation 4: confirmed locally. In `runner.yml`, the build renames the binary to `runner-${RUNNER_VERSION}-linux-{x86-64,arm-64}` (with `VERSION=canary` on cloud pushes), writes the `sha256sum` companion next to it, and uploads both files under an artifact of that same name, which matches the `runner-canary-linux-*` download pattern in `publish_runner_canary`. Observations 1 through 3 need no change, as noted in the review itself.

PR #942, opened 2026-07-15
Support updating a pinned plot in place

epompeii:
## Summary

Today you cannot meaningfully update a pinned plot. The `PATCH /v0/projects/{project}/plots/{plot}` endpoint existed but could only change `title`, `window`, and `index`, so to change what a plot actually shows (branches, testbeds, benchmarks, measures, x-axis, or the value/boundary toggles) you had to open it in the Perf explorer, re-pin a new plot, and delete the old one.

This extends the update path across every layer so an existing pinned plot can be edited in place.

## Changes

**Backend** (all new fields optional; `None` = leave unchanged, so existing `index`/`title`/`window` callers are untouched):
- `bencher_json`: add the value/boundary toggles, `x_axis`, and `branches`/`testbeds`/`benchmarks`/`measures` to `JsonPlotPatch`/`JsonPlotPatchNull` and the hand-written `Deserialize`.
- `bencher_schema`: add `QueryPlot::update`, which resolves the provided component UUIDs scoped to the project, then applies the scalar changeset, any rank move (including redistribution), and each provided component list replacement (delete + re-insert) atomically in a single write transaction via a unit-testable `apply_update` helper. A provided list replaces the current one wholesale.
- Component list validation, shared by `POST` and `PATCH`: a UUID that does not belong to the project is rejected with a `404 Not Found`, an empty list is rejected with a `400 Bad Request` (a plot missing any dimension would never render anything), and duplicate UUIDs are deduped preserving first-occurrence order. Previously `POST /v0/projects/{project}/plots` accepted foreign UUIDs and empty lists.
- `api_projects`: `patch_inner` delegates to `QueryPlot::update`.
- Regenerate `services/api/openapi.json`.

**CLI**: `bencher plot update` gains `--lower-value/--upper-value/--lower-boundary/--upper-boundary`, `--x-axis`, and `--branches/--testbeds/--benchmarks/--measures`.

**Console**: the Perf explorer's edit handlers no longer drop the loaded-plot association, so an "Update Pin" button (new `UpdateModal`) appears whenever a pinned plot is loaded; it saves the current selections back to that plot and presents the plot's current title, time window, and position for editing in place (disabled while any dimension is empty). The Pinned tab shows the loaded plot above its search box with an `x` to deselect it, matching the other dimension tabs. "Clear" still gives a fresh view for creating a brand-new plot.

Also adds the "Less is more / KISS" core axiom to `CLAUDE.md`.

## Testing

- `bencher_json`: 7 deserialization tests (absent vs null title, flags, x-axis, component lists, empty list, duplicate-field error).
- `bencher_schema`: `apply_update` tests (replace only provided components, scalar flags + x-axis) and rank tests (redistribution, rollback with the enclosing transaction).
- `api_projects`: 6 HTTP integration tests (full-config PATCH, unknown component 404, null title clear, index + components combined, empty list 400 on POST and PATCH, duplicate UUID dedupe).
- `cargo clippy --no-deps --all-targets --all-features -- -Dwarnings` clean, `cargo fmt --check` clean, `cargo check --no-default-features` passes.
- Console: Biome clean, `vitest` green; `bencher plot update --help` shows the new flags.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -110,6 +110,8 @@ The clippy script will install the target automatically and warn if no cross-com
 
 ## Code Quality
 
+**Core Axiom: Less is more. Err on the side of simplicity. KISS.**
+
 - Always run `cargo fmt` and `cargo clippy` when testing or before committing
 - Run `cargo fmt` one final time after all changes are complete (including any generated code or lint fixes), since clippy fixes and other automated changes can introduce formatting drift
 - Use `#[expect(...)]` instead of `#[allow(...)]` for lint suppression
--- a/lib/api_projects/src/plots.rs
+++ b/lib/api_projects/src/plots.rs
@@ -14,7 +14,7 @@ use bencher_schema::{
     model::{
         project::{
             QueryProject,
-            plot::{InsertPlot, QueryPlot, UpdatePlot},
+            plot::{InsertPlot, QueryPlot},
         },
         user::{
             actor::{ApiActor, PubProjectBearerToken},
@@ -367,12 +367,9 @@ async fn patch_inner(
     let query_plot =
         QueryPlot::get_with_uuid(auth_conn!(context), &query_project, path_params.plot)?;
 
-    let update_plot =
-        UpdatePlot::from_json(context, &query_project, &query_plot, json_plot.clone()).await?;
-    diesel::update(schema::plot::table.filter(schema::plot::id.eq(query_plot.id)))
-        .set(&update_plot)
-        .execute(write_conn!(context))
-        .map_err(resource_conflict_err!(Plot, (&query_plot, &json_plot)))?;
+    query_plot
+        .update(context, &query_project, json_plot)
+        .await?;
 
     auth_conn!(context, |conn| {
         QueryPlot::get_with_uuid(conn, &query_project, path_params.plot)
--- a/lib/api_projects/tests/projects.rs
+++ b/lib/api_projects/tests/projects.rs
@@ -698,3 +698,234 @@ async fn projects_soft_delete_endpoints_inaccessible() {
         );
     }
 }
+
+// Plot dimensions seeded directly in the DB for the plot PATCH test.
+struct PlotDimensions {
+    branch1: bencher_json::BranchUuid,
+    branch2: bencher_json::BranchUuid,
+    testbed: bencher_json::TestbedUuid,
+    benchmark: bencher_json::BenchmarkUuid,
+    measure: bencher_json::MeasureUuid,
+}
+
+#[expect(clippy::expect_used, reason = "test helper seeding plot dimensions")]
+fn seed_plot_dimensions(server: &TestServer, project_id: i32) -> PlotDimensions {
+    use bencher_api_tests::helpers::base_timestamp;
+    use bencher_json::{BenchmarkUuid, BranchUuid, MeasureUuid, TestbedUuid};
+    use bencher_schema::schema;
+    use diesel::{ExpressionMethods as _, RunQueryDsl as _};
+
+    let now = base_timestamp();
+    let branch1 = BranchUuid::new();
+    let branch2 = BranchUuid::new();
+    let testbed = TestbedUuid::new();
+    let benchmark = BenchmarkUuid::new();
+    let measure = MeasureUuid::new();
+
+    let mut conn = server.db_conn();
+    for (uuid, name) in [(&branch1, "branch-one"), (&branch2, "branch-two")] {
+        diesel::insert_into(schema::branch::table)
+            .values((
+                schema::branch::uuid.eq(uuid),
+                schema::branch::project_id.eq(project_id),
+                schema::branch::name.eq(name),
+                schema::branch::slug.eq(&format!("{name}-{uuid}")),
+                schema::branch::created.eq(&now),
+                schema::branch::modified.eq(&now),
+            ))
+            .execute(&mut conn)
+            .expect("Failed to insert branch");
+    }
+    diesel::insert_into(schema::testbed::table)
+        .values((
+            schema::testbed::uuid.eq(&testbed),
+            schema::testbed::project_id.eq(project_id),
+            schema::testbed::name.eq("testbed-one"),
+            schema::testbed::slug.eq(&format!("testbed-one-{testbed}")),
+            schema::testbed::created.eq(&now),
+            schema::testbed::modified.eq(&now),
+        ))
+        .execute(&mut conn)
+        .expect("Failed to insert testbed");
+    diesel::insert_into(schema::benchmark::table)
+        .values((
+            schema::benchmark::uuid.eq(&benchmark),
+            schema::benchmark::project_id.eq(project_id),
+            schema::benchmark::name.eq("benchmark-one"),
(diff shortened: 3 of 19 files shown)


epompeii:
Thanks for the review. Addressed in `29864793`:

**Fixed**
- **Items 1 + 2 (project scoping + 500→404)**: Both create (`InsertPlot::from_json`) and update (`QueryPlot::update`) now resolve component UUIDs with the project-scoped `QueryBranch::from_uuid(conn, query_project.id, uuid)` (and the testbed/benchmark/measure equivalents) instead of the UUID-only `get_id`. A foreign or non-existent UUID is now rejected with a `resource_not_found_err!` (404) up front rather than silently referenced or surfaced as a 500. Added an integration test (`plot_patch_rejects_unknown_component`) asserting the 404.
- **Item 4c (dead code)**: Pruned the now-unused `InsertPlot{Branch,Testbed,Benchmark,Measure}::from_json` variants (superseded by `from_resolved`) and their orphaned imports.

**Deliberately skipped (with rationale)**
- **Item 3 (rank recompute in a separate transaction)**: Pre-existing, and it only matters when `index` triggers a redistribution *and* the component write then fails; the console "Update Pin" flow sends no `index`, so this path isn't exercised there. Folding it in means reworking `update_rank`'s `immediate_transaction` (which doesn't nest) and touches create too, so it's out of scope for this PR.
- **Item 4a (one connection for the resolve loop)**: Kept the per-UUID checkout so create and update stay consistent; it's a micro-optimization and the simpler diff.
- **Item 4b (`onKeyDown` closes on any key)**: Matches the existing `PinModal`/`ShareModal` behavior; changing only this modal would be inconsistent.

Full workspace clippy `-Dwarnings`, `cargo fmt --check`, and the plot unit + integration tests are green.

epompeii:
Thanks, that's a thorough second pass. Deliberate decisions on the remaining items (all pre-existing patterns or degenerate cases, and consistent with the create path):

- **#1 rank atomicity**: Leaving as-is. `update_rank`'s redistribution uses `immediate_transaction`, which can't nest inside `write_transaction!` (per CLAUDE.md), so folding it in means reworking `update_rank`/`new_rank` (which also affects create). The window is narrow (only when `index` triggers a redistribution *and* `apply_update` then fails), the "Update Pin" console flow sends no `index`, and the fallout is plot ordering, not data loss. Out of scope for this PR; worth a dedicated follow-up if we want create + update fully atomic.
- **#2 empty list clears**: Kept consistent with create, which also does not enforce a non-empty floor. The console only ever sends real selections, and the Perf renderer already handles a zero-series plot (reachable today via create). Enforcing a non-empty invariant would be a behavior change to create too, so I'd rather do that deliberately in its own change.
- **#3 per-UUID connection**: Kept per-UUID so create and update stay identical; it's a micro-optimization.
- **CLI can't clear a list / `UpdateModal` closes on any key / `too_many_lines` expects**: clearing to empty is a degenerate op the UI never needs; the modal matches the existing `PinModal`/`ShareModal` behavior (changing only this one would be inconsistent); and the two `too_many_lines` expects are inherent per-field boilerplate (12 fields x null-vs-absent handling), not avoidable complexity.

`cargo gen-types` was run (openapi regenerated, `bencher.ts` unchanged since no typeshared type changed), plus clippy `--all-features -Dwarnings`, `--no-default-features` check, `cargo fmt --check`, and the plot unit/integration tests (`bencher_json --features server`, `bencher_schema --features plus`, `api_projects`).

epompeii:
Thanks for the second pass. Addressed in `588251ef`:

**Fixed**
- **#3 (clock)**: `QueryPlot::update` now stamps `modified` with `context.clock.now()` instead of `DateTime::now()`. The `DateTime::now()` calls inside `new_rank`/`update_rank` remain since those helpers only receive a connection; they are pre-existing and shared with create.
- **#1 (UpdateModal guard comment)**: The guard itself was already correct; `isPlotInit()` is defined in `PerfPanel` as exactly "any of the branches, testbeds, benchmarks, or measures lists is empty", computed from the same memos whose values the modal sends. So deselecting all branches hides the "Update Pin" button (it renders inside `Show when={!isPlotInit()}`), disables the modal's Update button, and short-circuits `handleSubmit`; the `branches: []` request cannot be sent. The comment invited the misreading though, so it now spells out what `isPlotInit` means.

**No action (with rationale)**
- **#2 (RollbackTransaction detail)**: Keeping the documented tradeoff; the case is unreachable after a redistribution and the 409 status is unchanged.
- **#4 (TOCTOU)**: Pre-existing pattern shared with create; a component deleted between resolution and insert fails the FK inside the transaction and maps to a 409.

Also in this push (follow-up feature): the Pinned tab now shows the loaded plot above its search box with an `x` to deselect it, matching the selected rows on the dimension tabs. The `x` only nulls the `plot` param; the rest of the view is untouched, so the "Pin" button reappears for pinning the edited view as a new plot.

Verification: workspace clippy `--all-features -Dwarnings` clean, `cargo fmt --check` clean, `bencher_schema` and `api_projects` suites green, Biome format + lint clean, `vitest` green.

epompeii:
Thanks. Addressed in `63d187ef`:

**Fixed**
- **#1 (clock consistency)**: `update_rank` now takes the caller's resolved `modified` timestamp instead of calling `DateTime::now()` in the redistribution loop, so a single update uses a single clock and the rank path is deterministic in tests. The `new_rank` create path keeps its pre-existing behavior; that is a separate change if we want it.

**Confirmed intent (no change)**
- **#3 (POST behavior changes)**: Deliberate. Rejecting empty dimension lists was an explicit product decision (a plot missing any dimension never renders), and the cross-project 404 plus dedupe are covered by tests and the changelog.
- **#4 (Update Pin scope)**: Deliberate. Saving the current Perf view's selections back to the plot is the point of the feature; the modal copy states it, and the title/window/position fields were added so those can be edited in the same action.
- **#5 (PLOT_PARAM retention)**: Correct, that is the central behavioral shift.

**Skipped (with rationale)**
- **#2 (macro vs fn)**: A plain fn does not work here: `auth_conn!` awaits per UUID (no async closures) and `from_uuid` is an inherent method per query type. Also the two remaining `too_many_lines` expects are on the hand-written `JsonUpdatePlot` deserializer, not on anything the macro forces; `QueryPlot::update` itself no longer carries one.
- **#6 (second plots-list fetch)**: `plots_tab` is paginated, so the panel does not hold the full list. One bounded fetch (64-plot cap) when the modal opens is simpler than threading a new full-list resource through the panel.

Verification: workspace clippy `--all-features -Dwarnings` clean, `bencher_schema` 170/170, `api_projects` 178/178, Biome + vitest green.

epompeii:
All five points are acknowledged as no-action for this PR:

- **#1 (`new_rank` clock + transaction)**: Agreed, follow-up material; it needs `context` threading and touches the create path.
- **#2 (per-UUID conn) and #3 (full rewrite on save)**: Known tradeoffs, kept for consistency and simplicity at a 64-plot/64-component scale.
- **#4 (window NaN)**: Unreachable; `JsonPlot.window` is a required `u32` and the submit button is disabled until the form is seeded from it.
- **#5 (PLOT_PARAM retention)**: Intended product behavior; the pin persists while the view diverges so it can be updated in place or deselected via the Pinned tab `x`.

CI is green on `63d187ef`: 52 checks passed, 15 expected skips.

PR #946, opened 2026-07-16
Cancel Stripe subscription when an organization is deleted

epompeii:
## Problem

When an organization has a paid Stripe subscription, deleting the organization left the subscription live in Stripe, so the customer kept being billed for an organization that no longer exists. Organization deletion (`delete_inner`) never touched the `plan` table or Stripe in either the soft-delete or the admin hard-delete path.

## Fix

Cascade an immediate subscription cancellation on organization deletion, reusing the existing cancellation logic:

- New `cancel_and_remove_plan` helper (`lib/api_organizations/src/plan.rs`) cancels any live Stripe subscription and removes the local `plan` row. It is a no-op when there is no biller (Self-Hosted) or no plan row.
- `delete_inner` (`lib/api_organizations/src/organizations.rs`) calls it under the `plus` feature in both the soft-delete and admin hard-delete branches, before the organization is deleted. Cancel-first means a Stripe failure aborts the deletion, so a live subscription is never orphaned (the delete stays retryable).
- `delete_plan` is now status-aware for metered plans: it checks `metered_plan_status` and only calls `cancel_metered_subscription` when the subscription is still live (reusing `existing_plan_action`). An already-canceled subscription is skipped instead of erroring, which also makes the admin `DELETE .../plan` endpoint idempotent.

## Edge cases

- Already-canceled or lapsed subscription (stale row): skips Stripe, removes the row, deletion succeeds.
- Transient Stripe error: surfaces as 503, deletion aborts and is retryable, never orphans.
- Licensed / Self-Hosted plans: canceled and the organization license cleared via the existing `delete_plan` branch.
- No biller (Self-Hosted server): no-op.

## Testing

- `cargo nextest run -p api_organizations --features plus`: 40/40, including every deletion test (the new `#[cfg(feature = "plus")]` path cleanly no-ops with no biller configured).
- `cargo clippy --all-targets --all-features -Dwarnings`, `cargo check --no-default-features`, `cargo test --doc`, `cargo fmt --check`: all clean.
- OpenAPI spec regenerated (`cargo gen-types`); only the organization-delete endpoint description changed.

The full end-to-end "delete org cancels Stripe" path is not exercised in the standard suite because the API test harness uses `biller: None` with no mock Stripe. The decision logic (`existing_plan_action`) and the Stripe building blocks (the `TEST_BILLING_KEY`-gated `biller()` test) are covered. A biller mock for deterministic end-to-end coverage is a suggested follow-up.

### DIFF
--- a/lib/api_organizations/src/organizations.rs
+++ b/lib/api_organizations/src/organizations.rs
@@ -25,6 +25,9 @@ use dropshot::{HttpError, Path, Query, RequestContext, TypedBody, endpoint};
 use schemars::JsonSchema;
 use serde::Deserialize;
 
+#[cfg(feature = "plus")]
+use crate::plan::cancel_and_remove_plan;
+
 pub type OrganizationsPagination = JsonPagination<OrganizationsSort>;
 
 #[derive(Debug, Clone, Copy, Default, Deserialize, JsonSchema)]
@@ -334,6 +337,7 @@ async fn patch_inner(
 /// The user must have `delete` permissions for the organization.
 /// By default, organizations are soft-deleted (along with their child projects).
 /// Set the `hard` query parameter to `true` to permanently delete the organization (requires server admin).
+/// On Bencher Cloud, deleting an organization also immediately cancels any active subscription.
 #[endpoint {
     method = DELETE,
     path =  "/v0/organizations/{organization}",
@@ -384,6 +388,12 @@ async fn delete_inner(
                 Organization,
                 &path_params.organization
             ))?;
+
+        // Cancel any active Stripe subscription and remove the plan row before deleting the
+        // organization so no dangling subscription is left in Stripe.
+        #[cfg(feature = "plus")]
+        cancel_and_remove_plan(context, &query_organization).await?;
+
         diesel::delete(
             schema::organization::table.filter(schema::organization::id.eq(query_organization.id)),
         )
@@ -398,6 +408,11 @@ async fn delete_inner(
             auth_user,
             Permission::Delete,
         )?;
+
+        // Cancel any active Stripe subscription and remove the plan row before deleting the
+        // organization so no dangling subscription is left in Stripe.
+        #[cfg(feature = "plus")]
+        cancel_and_remove_plan(context, &query_organization).await?;
 
         // Soft delete the organization and all its child projects
         let now = context.clock.now();
--- a/lib/api_organizations/src/plan.rs
+++ b/lib/api_organizations/src/plan.rs
@@ -5,7 +5,7 @@ use bencher_endpoint::{
     CorsResponse, Delete, Endpoint, Get, Patch, Post, ResponseCreated, ResponseDeleted, ResponseOk,
 };
 use bencher_json::{
-    DateTime, OrganizationResourceId, PlanStatus,
+    OrganizationResourceId, PlanStatus,
     organization::plan::{JsonNewPlan, JsonPlan, JsonUpdatePlan},
 };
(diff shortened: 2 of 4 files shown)


epompeii:
Thanks for the review. Addressed in `87d81e66`:

**Issue #1 (licensed asymmetry) - fixed.** The licensed branch in `delete_plan` is now status-aware, symmetric with the metered branch: it checks `licensed_plan_status` and only calls `cancel_licensed_subscription` when the subscription is still live (`existing_plan_action == Conflict`). Added `Biller::licensed_plan_status` (mirroring `metered_plan_status`, mapping a Stripe 404 to `None`) and factored the shared 404-to-`None` logic into a private `subscription_status` helper. An already-canceled or gone licensed subscription no longer aborts organization deletion.

**Issue #2 (503 on Stripe outage) - intentional.** Aborting the delete when the subscription status can't be resolved is the deliberate fail-safe: better to block deletion transiently than to delete the org and orphan a possibly-live subscription. It matches the `prune_or_conflict_existing_plan` precedent, and the delete is retryable once Stripe recovers.

**Issue #3 (non-atomic partial-failure window) - acknowledged.** A Stripe call can't sit inside the DB transaction, so cancel plus local-row-delete, then org-delete, is inherently two steps. As noted it is recoverable: a retry short-circuits on the now-missing plan row (and a gone/terminal subscription is skipped), then proceeds.

**Testing.** The `no biller -> Ok` early return is exercised by the existing integration tests (`organizations_delete`, `organizations_hard_delete_as_admin`) run under `--features plus` with `biller: None` (they pass). The `no plan row -> Ok` return needs a biller present with no plan row, which the harness can't set up without a Stripe mock; a biller mock for deterministic end-to-end coverage remains the suggested follow-up. The shared decision logic stays covered by `existing_plan_action_decides`.

epompeii:
Thanks, another good catch. Addressed in `69813289`:

**Finding #1 (swallowing all DB errors as "no plan") - fixed.** `cancel_and_remove_plan` now uses diesel `.optional()`, so only `NotFound` becomes the no-op and every other database error propagates (mapped to a conflict) and aborts the delete. A transient DB error can no longer let the org be deleted while its subscription is still live.

**Finding #2 (tests).** Added a deterministic unit test (`map_status_maps_subscription_status`) for the Stripe-status to `PlanStatus` mapping that feeds `metered_plan_status`/`licensed_plan_status` and therefore the cancel-vs-skip decision. Together with the existing `existing_plan_action_decides`, the substantive new decision logic ("cancel only when live; skip when gone/terminal") is unit-covered.

The remaining paths you list (`cancel_and_remove_plan` past the biller guard, the cancel-then-remove and error-propagation flow, and `subscription_status`/`licensed_plan_status` 404-vs-error) all require a `Biller`, which the API test harness constructs as `None` and cannot build without live Stripe credentials (`Biller::new` resolves products against Stripe on startup). They are therefore unreachable in the deterministic suite; the `TEST_BILLING_KEY`-gated `biller()` test is where the real Stripe round-trips run. A `Biller` trait/mock to make the full cancel path unit-testable is the right follow-up, but a larger, separate refactor than this fix should carry.

**Observations.**
- Partial-failure window: added a doc comment on `cancel_and_remove_plan` explaining the non-atomic Stripe+DB boundary and why cancel-first is the safe ordering.
- Naming overlap: added doc cross-references between `licensed_plan_status` (`Option`, 404 to `None`) and `get_licensed_plan_status` (errors on 404).
- Soft delete removing the plan row: agreed there is no org restore path today, so this is intended; a live subscription must not survive the delete.

PR #947, opened 2026-07-16
Add a Y-Axis scale toggle to perf plots

epompeii:
## What

Add a three-mode Y-Axis scale to perf plots: Auto, Linear, and Log.

- **Auto** (default): preserves the existing adaptive power scaling
- **Linear**: a true linear scale that shows magnitudes as-is
- **Log**: a genuine logarithmic scale; it falls back to Auto when the data includes zero or negative values, since the logarithm is undefined there

The selection is:
- Toggled from the perf plot header button, cycling Auto, Linear, Log
- Saved in the plot URL and restored for pinned plots
- Settable via the API (`y_axis` on plot create and update) and CLI (`bencher plot create --y-axis`, `bencher plot update --y-axis`)
- Carried along by the Perf explorer "Update Pin" button with the rest of the display settings

Also includes two related Perf explorer fixes:
- Selecting a pinned plot from the Plots tab now applies the plot's saved X-Axis and Y-Axis instead of the defaults
- The Branch Version Number x-axis tick labels are thinned to fit the plot width; previously every version rendered a label, which overlapped into an unreadable smear on plots with many versions

## How

- New `YAxis` enum (`auto` | `linear` | `log`) in `bencher_json`, stored on the `plot` table via a new migration as an integer column with strong Diesel `ToSql`/`FromSql` impls
- `#[serde(default)]` keeps existing API clients and pinned plots working: an omitted `y_axis` means `auto`, matching prior behavior
- `y_axis` is included in `JsonPlotPatch`/`JsonPlotPatchNull` and `UpdatePlot`, so `PATCH /v0/projects/{project}/plots/{plot}` can change it in place
- The d3 y-scale construction is extracted from `LinePlot.tsx` into `scale.ts` with unit tests covering all three modes and the zero/negative fallback
- The version x-axis is an ordinal point scale, and Observable Plot renders one tick per domain value on a point scale (its automatic tick thinning only applies to continuous scales). New `ticks.ts` thins the labels to an even stride targeting one label per 80px of plot width, matching the `tickSpacing` continuous scales already use; every tick is a real domain value, as a point scale requires

## Testing

- `cargo nextest run -p bencher_json --features server` and `-p bencher_schema --features plus`
- Console `npx vitest run`, including new `scale.test.ts` and `ticks.test.ts`
- Verified against Observable Plot 0.6.16 in a DOM render that a point scale renders every domain value by default (256 labels) and that the explicit `ticks` array thins them as expected
- `cargo clippy --no-deps --all-targets --all-features -- -Dwarnings`, `cargo check --no-default-features`, `cargo gen-types`, Biome CI lint/format commands


epompeii:
Regarding the `Log` degenerate-domain observation (`min == max > 0`): verified against d3 directly. `d3.scaleLog().domain([5, 5]).nice()` expands the collapsed domain out to the surrounding powers of ten (`[1, 10]`), so the scale stays finite and renders with real ticks. A `min < max` guard falling back to Auto would actually render worse: the pow scale leaves the collapsed domain as `[5, 5]` and maps every value to the midpoint. So the guard is omitted and the behavior is now pinned by a regression test in `scale.test.ts` ("Log handles a degenerate domain where min equals max").

PR #950, opened 2026-07-17
Create the GitHub Check as in progress before benchmarks run

epompeii:
## Summary

`bencher run --github-actions` currently creates the `Bencher Report` GitHub Check only after the benchmark completes, always with a terminal conclusion. On a rerun, the stale conclusion from the previous run lingers until the new benchmark finishes, which can take hours for slow benchmarks, and a cancelled job leaves the stale conclusion forever.

This PR creates the check with an `in_progress` status before the benchmark command runs and then completes that same check run with the results once they finish:

- `GitHubActions::start_check` creates the in progress check before the benchmark runs (best effort: failures log a warning and never abort the run)
- The check run ID is carried in a `CheckRunHandle` and threaded down to the existing CI completion step, which now updates the check run in place instead of creating a new one
- If the in progress check could not be created, completion falls back to the previous behavior of creating the check with a conclusion
- If `bencher run` exits with an error before the results are posted (harness failure without `--allow-failure`, report send failure, remote job poll error or timeout), the check is completed as `failure` with a short summary instead of lingering in progress
- `--dry-run` never creates a check, same as before
- The handle is consumed with `Option::take` at the completion site, so the error path never overwrites a check that was already completed with the full report (for example the `--error-on-alert` path)

## Notes

- A hard kill (CI job cancellation, SIGKILL) still leaves the check in progress, since no code gets to run. That is strictly better than showing a wrong terminal conclusion, and a signal handler could be a follow up.
- The `--detach` path keeps its existing semantics: the check is completed at submit time with the empty results report; it is just more visible now because the in progress check appears first.
- octocrab's `retry` feature transparently retries 5xx responses on the in progress create, so transient GitHub hiccups do not even need the fallback path.

## Testing

- New unit tests for head SHA resolution, conclusion mapping, and check run output shape
- Verified end to end against a local API server and a request recording GitHub API stub across four scenarios: successful run (POST `in_progress` lands before the benchmark command executes, then PATCH with `conclusion: success` and the report summary), failing benchmark (PATCH with `conclusion: failure`), `--dry-run` (zero GitHub requests), and a rejected in progress create (falls back to POST with a conclusion)
- `cargo nextest run -p bencher_cli`, `cargo clippy --all-targets --all-features -- -Dwarnings`, `cargo check --no-default-features`, and `cargo fmt` are all clean

## Docs

- Documented the new behavior in the `--github-actions` section for all nine languages
- Added a changelog entry under `Pending`


epompeii:
Addressed the review findings in the latest push (5a1932da):

**Finding 1 (orphaned check on the `generate_report() == None` path): fixed.** The error hook in `exec_inner` no longer gates on `result.is_err()`. Every path that posts results consumes the check handle, so any unconsumed handle at the end of the run now marks the check as failed. This makes the invariant total instead of depending on that branch staying unreachable.

**Finding 2 (failed remote job completes the check from the fetched report): deferred.** This mirrors the pre-existing behavior, where the check was also created from whatever report was displayed. Making a failed or canceled remote job force a `failure` conclusion is a behavior change to remote job semantics that deserves its own issue rather than a ride-along here.

**Finding 3 (permission help text printed twice): declined.** The two warnings come from two distinct failed API calls (start and complete), and each message names the operation that failed. Suppressing the second would require threading state between the two call sites to dedupe a best-effort log line, which does not seem worth the complexity.

Also in this push: the in-progress and failed check summaries now render the same branded `Bencher Report` header as the completed report, so the check reads consistently in every state.

Local verification on the new commit: `cargo nextest run -p bencher_cli`, `cargo clippy --all-targets --all-features -- -Dwarnings`, and `cargo fmt` are all clean.

epompeii:
Addressed the second round of review feedback in 467986b5:

**Issue 1 (duplicate permission warning): partially addressed.** The start phase warning now notes that the check will still be created when the results are posted, so the second warning reads as the expected follow up rather than a mystery repeat. Full deduplication would require threading state between the two best effort call sites, which still does not seem worth it for a log line.

**Issue 2 (header markup duplication): fixed.** The shared title markup now lives in a single `pub const BENCHER_REPORT_TITLE` in `bencher_comment`, used by both `ReportComment::html_header` and the new CLI `branded_summary` helper, so the two cannot drift apart. Added unit tests for the in progress and failed summaries.

**Issue 3 (fail fallback invariant): addressed via comment.** The `exec_inner` comment now spells out that only error paths leave the handle unconsumed today, and that any future `Ok` path without results would mark the check as failed, which is still better than an eternally pending check. Exercising the fallback in an integration test would need an HTTP mock dependency, which this workspace intentionally avoids.

Local verification: `cargo nextest run -p bencher_cli -p bencher_comment` (35 tests), `cargo clippy --all-targets --all-features -- -Dwarnings`, `cargo check --no-default-features`, and `cargo fmt` are all clean.

PR #951, opened 2026-07-17
Generate a final invoice for metered usage on subscription cancel

epompeii:
## Problem

`Biller::cancel_metered_subscription` sent a bare `CancelSubscription` request. Stripe's `invoice_now` flag defaults to `false`, so an immediate cancel silently discarded any metered usage accrued since the last invoice. The customer was never billed for their final usage on either path that cancels a metered subscription: the admin plan DELETE endpoint and organization deletion.

## Fix

`cancel_metered_subscription` now cancels with `invoice_now(true)`, so Stripe generates a final invoice for any un-invoiced metered usage on the way out.

The licensed cancel path is unchanged: licensed plans bill upfront, so there is no un-invoiced usage to lose. Prorations stay off (`prorate` defaults to `false`) on both paths, so bare metal usage is billed in full and never prorated.

## Tests

The `metered_subscription` lifecycle test records usage before exercising the DELETE path. It now captures the creation-time invoice id and asserts that the cancel returns a subscription whose `latest_invoice` is present and different, verifying that the final invoice was generated. The `resubscribe_has_no_trial` helper cancels with no recorded usage, so the gated Stripe test also covers the nothing-to-bill case.

Only the invoice ids are held across await points in the test; keeping the full `Subscription` objects alive tripped `clippy::large_futures`.

### DIFF
--- a/plus/bencher_billing/src/biller.rs
+++ b/plus/bencher_billing/src/biller.rs
@@ -982,13 +982,19 @@ impl Biller {
     }
 
     /// Immediately cancel a metered subscription. Used by the admin-only DELETE
-    /// path; the organization downgrades to Free right away.
+    /// path; the organization downgrades to Free right away. `invoice_now`
+    /// generates a final invoice for any un-invoiced metered usage; a bare
+    /// cancel would silently discard all usage accrued since the last invoice.
     pub async fn cancel_metered_subscription(
         &self,
         metered_plan_id: &MeteredPlanId,
     ) -> Result<Subscription, BillingError> {
         let subscription_id: SubscriptionId = metered_plan_id.as_ref().into();
-        self.cancel_subscription(&subscription_id).await
+        CancelSubscription::new(subscription_id)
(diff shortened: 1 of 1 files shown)


epompeii:
Addressed the review's actionable item (test assertion could pass trivially when the creation subscription had no invoice) in b895b0ba.

Instead of comparing the final invoice against the *creation* invoice, the test now snapshots the latest invoice immediately before cancellation (reusing the already-fetched `resumed` subscription, so no extra Stripe call) and asserts the final invoice is distinct. Combined with the existing `.expect(...)`, this proves a new final invoice is generated in both cases: distinct from the prior invoice when one exists, and present-therefore-new when none did.

I did not add the literal `assert!(create_invoice_id.is_some())` suggestion because a Pro subscription starts in a native free trial and may not carry a creation-time invoice, which would make that assertion flaky.

The other two notes (org-delete billing side effect, customer-deletion ordering) are out of scope for this diff, as flagged.

PR #953, opened 2026-07-18
Add Bencher MCP server

epompeii:
Add `plus/api_mcp`, a stateless MCP (Model Context Protocol) server hosted by the API server at `POST /mcp`. On Bencher Cloud it will be fronted by `mcp.bencher.dev` (a DNS alias for the API server, like `registry.bencher.dev`), so the connector URL is `https://mcp.bencher.dev/mcp`. Self-Hosted servers serve it at `/mcp` on the API host.

## Design

- **Tool surface**: 36 tools, one per REST operation available to a project API key (`bencher_run_*`): project-scoped reads and non-destructive writes. No deletes, no renames, nothing outside a single project. A connected agent can never do more than the credential it holds.
- **Zero-drift dispatch**: tool handlers call the same (newly `pub`) `*_inner` functions as the REST endpoints in `api_projects` and `api_run`, so authorization semantics (project scoping, rename guards, status-only alert updates, per-actor rate limiting) are shared code, not a parallel implementation. A sync rule is recorded in `services/api/CLAUDE.md`.
- **Auth**: the same `Authorization: Bearer` credentials as the REST API (project key, user key, or token) via `ApiActor`; anonymous requests may read public projects; invalid credentials fail with `401` like REST. OAuth is a follow-up.
- **Transport**: stateless Streamable HTTP using the official `rmcp` SDK model types (`default-features = false`, no transport stack) with a custom dropshot handler: one JSON-RPC message per POST, `application/json` responses, notifications get `202 Accepted`, batches rejected, protocol versions 2025-03-26 through 2025-11-25 negotiated.
- **OpenAPI**: the MCP endpoints are `unpublished`; `openapi.json` is byte-identical, so there is no progenitor or gen-types churn.
- **`submit_run`**: accepts raw benchmark harness output and adapter settings; the server parses results and creates branches, testbeds, and thresholds on the fly, so agents can report results with no CLI installed.

## Docs

- New docs page at `/docs/how-to/mcp/` in all 9 languages, with shared code chunks; the CLI remains the recommended route when it is installed.
- The Claude plugin skill documents MCP alongside the CLI (`skills/bencher/mcp.md`).
- Changelog entry added.

## Testing

- 3 unit tests (protocol version negotiation, capabilities, tool registry invariants).
- 16 integration tests against a real in-process server: initialize, tools/list shape, batch rejection, method-not-found, unknown tool, invalid credential `401`, user token project listing, project key scoping (`403` cross-project), create-but-not-rename guard, anonymous public read, malformed arguments, and an end-to-end `submit_run` with a project key followed by `list_reports`.
- `cargo clippy --no-deps --all-targets --all-features -- -Dwarnings`, `cargo check --no-default-features`, `cargo deny check` (new `rmcp` dependency), and `cargo gen-spec` (spec unchanged) all pass.

## Follow-ups (not in this PR)

- Fly DNS and cert alias for `mcp.bencher.dev`.
- OAuth support for MCP clients that only support OAuth for remote servers.

Note for review: `handle_request` in `plus/api_mcp/src/server.rs` carries an `#[expect(clippy::wildcard_enum_match_arm)]` (precedent: `plus/bencher_oci/src/image.rs`), since enumerating all unsupported rmcp request variants would break on every rmcp upgrade.


epompeii:
Addressed the review observations in the latest push:

1. CORS on error responses: the shared `bencher_schema` error helpers already wrap every `HttpError` (401, 403, 429, etc.) with CORS headers via `cors_headers`, so auth and rate limit errors were covered. The only gaps were the two practically unreachable response serialization branches in `endpoint.rs`; those now use `issue_error`, which adds CORS headers and Sentry capture.
2. Corrected the `call_tool` comment to note that `list_projects` mirrors `GET /v0/projects`, which has no per-request rate limit.
3. Reworded the `McpTool` doc comment: a project API key is confined to its own project, while a user API key spans the user's projects, exactly as over REST.
4. Clarified the root CLAUDE.md Docker section: `services/api/Dockerfile` copies the whole workspace, while the other three Dockerfiles need a `cargo init --lib` stub per new crate.

epompeii:
Round two feedback addressed:

1. Rate limit consistency: the public rate limiter now also runs on the notification/response/error branch (previously an unmetered `202 Accepted`) and on the unknown-tool branch of `tools/call` (previously returned `invalid_params` before any limiter). The parse error and batch rejection paths were already covered.
2. Added a comment on `input_schema` noting that generation cannot fail for the concrete tool input types and that the `tool_specs_have_object_schemas` test guards the empty-object fallback.
3. Parse error responses with `id: null` are spec-compliant for unparseable requests, as noted; no change.

epompeii:
Round three feedback addressed:

1. Anonymous tool calls now hit the public rate limiter before dispatch, closing the valid-tool-with-malformed-arguments path. Note this makes MCP slightly stricter than REST, where dropshot rejects malformed bodies and query params with a `400` before any handler rate limiting runs. Authenticated bad-args requests remain pre-handler `400`s, matching REST parity.
2. `negotiate_protocol_version` now returns `SUPPORTED_PROTOCOL_VERSIONS[0]` instead of `ProtocolVersion::LATEST`, so an rmcp upgrade cannot advertise a version outside the supported set; the unit test asserts the lockstep.
3. Added a comment on `McpTool::ALL` warning that a variant missing from the array silently disappears from `tools/list`, unlike the compiler-checked matches.
4. The server instructions now document the list envelope: items under `data`, total under `total_count`.
5. `cargo deny check` passes locally (advisories, bans, licenses, sources) and the `Lint / Cargo Deny` and `Lint / Cargo Deny Advisories` CI jobs are green on this PR.

On the perf nit: `perf_image` uses the exact `JsonPerfImgQueryParams` to `JsonPerfQueryParams` conversion the REST `/v0/projects/{project}/perf/img` endpoint uses, so field preservation is shared code.

epompeii:
Round four feedback addressed:

1. `McpTool::ALL` is gone: the enum now derives `strum::EnumIter` (added `strum` as a workspace dependency; it was already in the tree transitively), so the complete tool list is derived and a new variant cannot be silently omitted from `tools/list`. The remaining `name`/`spec`/`try_call` matches are all compiler-checked exhaustive. `cargo deny check` still passes.
2. Origin validation: documented the conscious decision on `mcp_cors_headers`: bearer-only auth with no cookie or ambient credentials and `Access-Control-Allow-Origin: *` by design means DNS rebinding grants nothing beyond what any cross-origin client already has.
3. Added the "tool input types must not be recursive" invariant to the `input_schema` docs (inlining recurses without bound).
4. CORS split between `mcp_options` (shared `Endpoint::cors`) and the manual POST headers mirrors the `api_oci` pattern; left as is for consistency with the sibling crate.
5. Softened the `SUPPORTED_PROTOCOL_VERSIONS` comment: the transport was introduced in 2025-03-26; older clients are offered the newest supported version and may disconnect, per spec.

epompeii:
Round five feedback addressed:

1. `tools/list` is now cached: the static tool list is generated once behind a `LazyLock` and cloned per request (the clones are cheap: static name and description `Cow`s plus `Arc` schemas).
2. Response duplication (text content plus `structuredContent`) is the MCP compatibility recommendation and is kept as is; clients that predate structured content read the text block.
3. `cargo deny check` passes locally with both `rmcp` and `strum`, and the `Lint / Cargo Deny` and `Lint / Cargo Deny Advisories` CI jobs are green.
4. The non-recursive tool input type invariant is now recorded in the `services/api/CLAUDE.md` MCP section, next to the sync rule.
5. The two error channels (HTTP `401` for credentials, `isError` tool results for `400`/`403`/`404`) are now documented explicitly in the skill protocol notes.

epompeii:
Round six feedback addressed:

1. Added the requested tradeoff comment on `success_result`: the payload is deliberately carried as both text and structured content per the MCP compatibility recommendation, with a note to revisit if response sizes become a problem.
2. Credential brute-forcing observability: `/mcp` uses the same `ApiActor` extractor as REST, so failed project key authentication increments the existing OTel adverse event counters (`ProjectKeyAuthFailed` in `bencher_schema` `actor.rs`) and invalid user key and token paths flow through the same shared auth code. The exposure and protections are identical to every REST endpoint.
3. Verification confirmed locally: `cargo check -p api_mcp --features plus`, workspace `cargo clippy --no-deps --all-targets --all-features -- -Dwarnings`, `cargo check --no-default-features`, `cargo nextest run -p api_mcp --features plus` (20 tests), `cargo test --doc`, `cargo deny check`, and `cargo gen-spec` leaves `openapi.json` byte identical.

epompeii:
Round seven feedback addressed:

1. `cargo deny check` confirmation: it passes locally with both `rmcp` and `strum` (advisories, bans, licenses, sources), and the `Lint / Cargo Deny` and `Lint / Cargo Deny Advisories` CI jobs are green on every push of this PR. As the review notes, `default-features = false` keeps the `rmcp` transitive footprint to just the model types.
2. Added the suggested guard: `success_result` now carries a `debug_assert!(value.is_object())` with a comment, so a future tool returning a bare array or scalar is caught in debug and test builds before it can emit spec violating `structuredContent`.

The remaining observations (payload duplication watchpoint, authenticated malformed-args parity with REST, CORS rationale) are documented decisions carried over from earlier rounds.

epompeii:
On the CI filter question: no `ci.yml` change is needed. The `docker` filter watches the Dockerfile paths themselves (`services/api/Dockerfile`, `services/cli/Dockerfile`, `services/console/Dockerfile`, `docker/**`), and this PR modifies three of them, so the Docker jobs were triggered and are green (`Test / Docker API Smoke Test`, `Test / Bench Image`). The `api_mcp` crate content is covered by the `rust` filter's `plus/**` glob, and the `runner` filter correctly excludes it since it is not a runner crate.

The recommended pre-merge confirmations are all CI jobs on this PR and pass on every round: `Lint / Cargo Clippy` (workspace, `-Dwarnings`), `Test / Cargo Check Workspace (not(plus))` for the no-default-features build, and `Test / Cargo Test` for the `plus`-feature test run.

PR #954, opened 2026-07-18
Add `cargo ops isolate` subcommand for CPU isolation boot args

epompeii:
## What

Add a new `cargo ops isolate <runner>` subcommand to the runner ops tooling. It configures the kernel boot args for full CPU isolation on a bare metal runner server (`isolcpus=`, `nohz_full=`, `rcu_nocbs=`), then reboots the server and verifies the args took effect.

- Idempotent: exits early if `/proc/cmdline` already contains `isolcpus=` or `nohz_full=`, mirroring the runner preflight detection
- Benchmark CPU list defaults to `1-(nproc-1)` queried from the server (CPU 0 stays housekeeping); `--cpus` overrides
- Writes a GRUB drop-in at `/etc/default/grub.d/99-bencher-isolation.cfg` that appends to `GRUB_CMDLINE_LINUX_DEFAULT`, then runs `update-grub`
- Stops the runner service, reboots, waits for SSH to come back, and verifies the new cmdline
- Reports whether the runner service came back up after the reboot

## Why

The runner preflight added in #923 warns when the kernel boots without `isolcpus=`/`nohz_full=`. The runtime cpuset partition only partially compensates: tickless cores (`nohz_full`) and RCU callback offload (`rcu_nocbs`) can only be configured at boot. This subcommand makes applying the boot args a repeatable operation instead of a manual SSH session.

## How

- `tasks/runner_ops/src/task/isolate.rs`: new task, with unit tests for the cmdline detection, CPU list derivation, and drop-in contents
- `tasks/runner_ops/src/parser/mod.rs`: new `isolate` subcommand parser
- `tasks/runner_ops/src/task/ssh.rs`: the reboot wait loop moved out of `install_os.rs` into a shared `Ssh::wait_for_reboot()` method
- `tasks/runner_ops/CLAUDE.md`: document the new operation

### DIFF
--- a/tasks/runner_ops/CLAUDE.md
+++ b/tasks/runner_ops/CLAUDE.md
@@ -51,6 +51,15 @@ cargo ops logs <runner> --follow
 cargo ops provision <runner>
 ```
 
+### CPU isolation boot args
+
+Configure `isolcpus=`/`nohz_full=`/`rcu_nocbs=` kernel boot args for the benchmark cores via a GRUB drop-in (`/etc/default/grub.d/99-bencher-isolation.cfg`), then reboot the server and verify. This clears the runner preflight notice about missing isolation boot args. Idempotent: exits early if the cmdline already has the args (presence-only; it will not re-scope an existing CPU list, even with `--cpus`). The benchmark CPU list defaults to `1-(nproc-1)` (CPU 0 is housekeeping); override with `--cpus`.
+
+```bash
+cargo ops isolate <runner>
+cargo ops isolate <runner> --cpus 1-5
+```
+
 ## How It Works
 
 1. `deploy` downloads the runner binary from a GitHub Actions CI artifact, SSHes into the server, stops the existing service, copies the new binary, configures systemd, and starts the service.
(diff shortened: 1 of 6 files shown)


epompeii:
Addressed the review feedback in 4cd123e:

1. Reboot race: `wait_for_reboot` is now boot ID aware. The boot ID is captured before `reboot`, and a probe only counts once `/proc/sys/kernel/random/boot_id` differs, so a not-yet-rebooted system can no longer satisfy the wait. `install_os` uses the same mechanism.
2. `--cpus` validation: the CPU list is validated (digits, `-` ranges, `,` separators, no empty parts) before anything is written, so typos fail fast instead of after a reboot cycle.
3. Added a doc note on `benchmark_cpus` about the contiguous CPU numbering assumption, pointing at `--cpus` for non-trivial topologies.

Kept the `stop_service` call before reboot as a graceful shutdown courtesy.

epompeii:
Addressed in 45fb93f:

1. `validate_cpu_list` now takes `nproc` (always queried) and bounds every value: CPU 0 is rejected as the housekeeping core and values must be below `nproc`, so out-of-range lists fail before the reboot cycle.
2. The not-active-yet hint now also points at `cargo ops start`.
3. Presence-only verification: left as is, documented behavior.
4. Cleaned up the pre-existing emdash comment in `install_os.rs`.

PR #955, opened 2026-07-18
Sort `cargo ops isolate` GRUB drop-in after provider drop-ins

epompeii:
## What

Follow-up to #954, fixing an issue found in field testing of `cargo ops isolate`:

- Rename the GRUB drop-in from `99-bencher-isolation.cfg` to `zz-bencher-isolation.cfg`. Provider images can ship their own `/etc/default/grub.d` drop-in (e.g. Hetzner's `hetzner.cfg`) that overwrites `GRUB_CMDLINE_LINUX_DEFAULT` instead of appending to it. Drop-ins are sourced in glob order, so `99-` sorted before `hetzner.cfg` and our args were silently clobbered; `zz-` sorts last.
- Verify the generated `/boot/grub/grub.cfg` actually contains the isolation args after `update-grub` and before rebooting. A clobbered cmdline now fails fast with a pointer at drop-in ordering, instead of only surfacing after a wasted reboot cycle.

## Why

Without the ordering fix, the command completed `update-grub` successfully but the rebooted kernel came up without any isolation args, and the failure was only detected by the post-reboot cmdline check.

## How

- `tasks/runner_ops/src/task/isolate.rs`: `GRUB_DROP_IN` renamed with a comment explaining the ordering constraint; pre-reboot `grep` check on the generated config
- `tasks/runner_ops/CLAUDE.md`: document the drop-in name and why it sorts last

### DIFF
--- a/tasks/runner_ops/CLAUDE.md
+++ b/tasks/runner_ops/CLAUDE.md
@@ -53,7 +53,7 @@ cargo ops provision <runner>
 
 ### CPU isolation boot args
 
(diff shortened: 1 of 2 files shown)


epompeii:
Took the optional suggestion: the pre-reboot check now uses `grep -qF` for an explicit fixed-string match.

Issue #778, opened 2026-04-09
gungraun: Fix benchmark name should not depend on the (truncated) description


gamma0987:
As mentioned in #638 the benchmark name in bencher uses the whole first line as benchmark name which consists of the module path the optional id and a description. However, a benchmark can be uniquely identified by the module path and if present, the id. Since the description is possibly truncated and isn't required to distinguish benchmarks, it shouldn't be part of the benchmark name.

epompeii:
@gamma0987 when was this description added?

My worry is that this is going to be a breaking change for all users of [the `rust_gungraun` adapter](https://bencher.dev/docs/explanation/adapters/#-rust-gungraun). Any change to the benchmark name will effectively create a new Benchmark with no historical data.

I see two potential paths forward:
1. Limit this change to https://github.com/bencherdev/bencher/issues/638 
2. Add `aliases` to Benchmarks, including in BMF

Do you see another option?

gamma0987:
I've been looking at the version history of `lib/bencher_adapter/src/adapters/rust/iai_callgrind.rs` and noticed the description has always been part of the benchmark name.

> Add aliases for benchmarks, also in BMF

I'm not sure how this is supposed to work, but a database migration would be an alternative. The new id can easily be derived from the existing benchmark name. This corresponds exactly to the logic I've already implemented in the adapter for this pull request.

> Limit this change to https://github.com/bencherdev/bencher/issues/638

I don't expect any significant differences between the two parsers and would therefore try to make the various parser options consistent. This way, users can easily switch between them.

Alternatively, everything could be left as is. The main problem, however, is that a new benchmark name and a new database entry are created as soon as someone makes a change to the benchmark settings that are irrelevant to the benchmark case but are shown in the description. The description was never meant as an identifier in contrast to `file::group::function id`. Using the id without a description would be more stable and convenient.

epompeii:
> I've been looking at the version history of `lib/bencher_adapter/src/adapters/rust/iai_callgrind.rs` and noticed the description has always been part of the benchmark name.

Thank you for doing the digging @gamma0987!

> > Add aliases for benchmarks, also in BMF
> 
> I'm not sure how this is supposed to work

It would complicate BMF quite a bit, but we could allow for an alias object or a metric object:

```json
{
  "new_benchmark": {
    "value": 42.0
  },
  "old_benchmark": {
    "alias": "new_benchmark"
  }
}
```

Or something like that. I haven't totally thought there this though. Just spitballing.

[...]

PR #962, opened 2026-08-08
Support gungraun json format

### LINKED ISSUE #638: Support gungraun json output
I initially tried uploading the json format, as I did not expect the plain console output to be parsed.

I think parsing the json has some advantages here:

- Its meant for machine parsing. I would not expect the human readable output to be stable
- Its easier to parse
- The benchmark name does not depend on the truncated name in the output. (id and function_name / module_path are available)
- More metrics available

gamma0987:
This PR adds a gungraun JSON v6 adapter to complement the existing text-based gungraun parser, enabling users to consume machine-readable output directly.

What changed:

- Fixes the missing DHAT metrics also for the text-based gungraun parser in `rust/gungraun.rs`.
- New `AdapterRustGungraunJson` in `lib/bencher_adapter/src/adapters/rust/gungraun_json.rs`
- Uses [gungraun-summary](https://crates.io/crates/gungraun-summary) to parse the JSON format. These are the types directly exported from gungraun. This eliminates the error source and additional work parsing the JSON manually in bencher.
- Uses `gungraun_summary::util::parse_slice()` for version-aware parsing (only v6 currently)
- Parses NDJSON input: one benchmark summary per line. This is the default gungraun JSON format. gungraun can print in pretty JSON, but parsing that format is not covered in this pr.
- Supports all 8 Gungraun tools with complete metric coverage. 2 tools (Massif and BBV) don't produce metrics and return an empty vector of GungraunMeasure(s). Massif and BBV return empty benchmark vectors, which afaik surface in the UI with a warning if they are the only tools.
- Rejects Mean/Median averaging like the text-based gungraun parser
- Registers a new `rust_gungraun_json` adapter in addition to the existing `rust_gungraun` adapter

Documentation updates are intentionally deferred to a follow-up PR.

Note on benchmark naming and ids:
Already discussed in #778. This parser implementation uses (`file::group::function` + `::id`) and the optional id as benchmark name to identify gungraun benchmarks uniquely. In addition to #778, I want to add that the main problem is that text parsed benchmark names include the "most likely" truncated `BenchmarkSummary::details`. Exactly replicating the text parsed benchmark name is impossible because the user can disable or change the truncation length with configuration values. The JSON summary doesn't reflect that decision and doesn't store the user provided truncation configuration value. Required follow-up PR: A database migration to convert the old benchmark names to the new ones. Changing the old names to the new names is straightforward but irreversible although I don't think there's any need to go back.

Closes #638


epompeii (changes requested):
@gamma0987 thank you for the contribution!

I'm thinking we should extend the existing tree structure of the adapters:

1. `rust_gungraun_json` is your new JSON adapter
2. `rust_gungraun_stdout` is the existing `nom` parser
3. `rust_gungraun` is now a parent node that attempts to parse JSON then stdout

This will be another level in the tree below: `magic` -> `rust` -> `rust_gungraun`.

gamma0987:
Sure. No problem

gamma0987:
What changed (CLI):

- `rust/gungraun.rs` is now just a dispatcher that tries `gungraun_json::AdapterRustGungraunJson::parse` then falls back to `gungraun_stdout::AdapterRustGungraunStdout::parse`.
- Bugfix in the same file: `from_sql` was missing for `RUST_GUNGRAUN_JSON_INT` in the prior push.

UI:

- Removed stale `Adapter.RustIaiCallgrind` references
- Added `RustGungraun`, `RustGungraunJson`, `RustGungraunStdout` to output surfaces (icon, command, ...)
- Fixed `gungraun.astro` Learn More href.

lib/bencher_json/src/project/report.rs
```diff
@@ -90,6 +90,8 @@ const RUST_BENCH_INT: i32 = 21;
 const RUST_CRITERION_INT: i32 = 22;
 const RUST_IAI_INT: i32 = 23;
 const RUST_GUNGRAUN_INT: i32 = 24;
+const RUST_GUNGRAUN_JSON_INT: i32 = 25;
+const RUST_GUNGRAUN_STDOUT_INT: i32 = 26;
```
epompeii: Pedantic but lets have `stdout` be first and `json` be second here. `stdout` is the default so IMO it make sense to list if first here and everywhere (other than the  actual `rust_gungraun` parsing, that is).

lib/bencher_adapter/tool_output/rust/gungraun/json_multiple_benchmarks.txt
```diff
@@ -0,0 +1,17 @@
...
...:null,"part":1,"path":"/callgrind.bench_play_game_100.out","pid":12345,"thread":1}},"metrics_summary":{"Callgrind":{"D1MissRate":{"diffs":{"diff_pct":"0","factor":"1"},"metrics":{"Left":{"Float":0.1}}},"D1mr":{"diffs":{"diff_pct":"0","factor":"1"},"metrics":{"Left":{"Int":6}}}}}}],"total":{"regressions":[],"summary":{"Callgrind":{"D1MissRate":{"diffs":{"diff_pct":"0","factor":"1"},"metrics":{"Left":{"Float":0.1}}},"D1mr":{"diffs":{"diff_pct":"0","factor":"1"},"metrics":{"Left":{"Int":6}}}}}}},"tool":"Callgrind"},{"flamegraphs":[],"log_paths":["/dhat.bench_play_game_100.log"],"out_paths":["/dhat.bench_play_game_100.out"],"summaries":{"parts":[{"details":{"Right":{"command":"/bencher/examples/rust/gungraun/target/release/deps/play_game-4a31ce0553c3b8ad --gungraun-run 00000 00000 00000","details":null,"parent_pid":null,"part":1,"path":"/dhat.bench_play_game_100.out","pid":12345,"thread":1}},"metrics_summary":{"Dhat":{"AtTEndBlocks":{"diffs":null,"metrics":{"Left":{"Int":1}}}}}}],"total":{"regressions":[],"summary":{"Dhat":{"AtTEndBlocks":{"diffs":null,"metrics":{"Left":{"Int":1}}}}}}},"tool":"DHAT"}],"project_root":"/bencher/examples/rust/gungraun","summary_output":null,"version":"6"}
```
epompeii, line 6: It seems like we are supporting [the `--output-format` option](https://gungraun.github.io/gungraun/latest/html/cli_and_env/basics.html?highlight=pretty-json#the-command-line-arguments) set to `json` but not `pretty-json`?
Would it be worth going ahead and covering both of them with the `rust_gungraun_json` adapter? I can see this being a footgun for Bencher users where they also want a somewhat human readable JSON output in a file that they then pass to `bencher`.

epompeii (commented):
Thank you for the updates @gamma0987.
Main question is on the `--output-format` support.

gamma0987:
Done. I'd thought the pretty JSON parser would be more work, which is why I originally deferred it to a follow-up PR. Turns out it wasn't so bad, so here it is :) I hope this is the direction you wanted.

epompeii (approved):
Thank you, @gamma0987!

PR #1002, opened 2026-08-23
Close the migration deploy window: restore the machine size, restart policy, and deploy wait

epompeii:
**Merge after the stack deploy completes and the API is healthy.**

This is the second half of a pair. The first half, #1001, sizes the API machine and the deploy wait for the migration window; this one puts both back. It is open now so that closing the window is a merge rather than a thing someone has to remember how to write at the end of a long night.

## What it restores

- `[[vm]]`: `cpu_kind = "shared"`, `cpus = 8`, `memory_mb = 2048`.
- `[[restart]]`: `policy = "on-failure"`, `retries = 6`.
- The prod deploy step: `--wait-timeout 300`.

Nothing here is a judgment about those values. They are the values the repository had before the window, and this returns them unchanged.

## Why the diff looks empty

This branch is the window commit plus its revert, so against `devel` today it is a no-op. Once #1001 merges, this branch's diff against `devel` becomes exactly the revert. That is the intended shape, not a mistake.

Merging this deploys once more with the restored settings, which is the deploy that leaves prod in its steady state.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -204,7 +204,7 @@ jobs:
       - uses: superfly/flyctl-actions/setup-flyctl@ed8efb33836e8b2096c7fd3ba1c8afe303ebbff1 # master
       - name: Deploy API to Fly.io Prod
         working-directory: ./services/api
-        run: flyctl deploy --local-only --config fly/fly.toml --wait-timeout 14400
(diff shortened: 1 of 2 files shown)


epompeii:
Closing: the deploy-window design was replaced by permanent operational tuning in #1001 (memory headroom, retry reduction, and a 3600s wait-timeout), so there is nothing left to revert.

PR #1010, opened 2026-08-24
Add the parameter resource endpoints

epompeii:
A parameter set becomes a first class resource with full parity to the other dimension entities. It nests under its benchmark and is addressed by UUID only, because it has neither a name nor a slug, the same way a report or an alert is addressed.

## Routes

Under `/v0/projects/{project}/benchmarks/{benchmark}/parameters`:

- `OPTIONS`, `GET` (list): pages and sorts like the benchmark list, defaulting to creation order, with `X-Total-Count` on the response. An `archived` filter carries the same two states a benchmark list's does. There is no search parameter: there is nothing to search by.
- `POST` (create): body `{ "set": { ... } }`.

Under `/v0/projects/{project}/benchmarks/{benchmark}/parameters/{parameter}`:

- `OPTIONS`, `GET` (view).
- `PATCH`: body `{ "archived": true | false }`, mirroring benchmark archive semantics exactly.
- `DELETE`.

Permissions mirror the benchmark endpoints: a public project's sets are readable without authentication, and create, edit, and delete each take the same permission the corresponding benchmark endpoint takes.

## Create is create, not get-or-create

A parameter set that already exists under the benchmark collides on `UNIQUE (benchmark_id, "set")` and comes back as a conflict, which is the same error class a duplicate benchmark name returns. Two spellings of one set are one set, so `{"size_mb": 16.0}` conflicts with `{"size_mb": 16}`, and posting `{}` conflicts with the empty set the benchmark was born with.

Because this path mints parameter sets, it carries the per project creation ceiling that report ingest already enforces, with the same error. Without it, this endpoint would be an unbounded way to mint rows, grid points, and billable series.

## Delete takes the results with it

Deleting a parameter set deletes every result that ran with it: the `report_benchmark` rows that carry it, the metrics and boundaries that cascade from those, and the billable series that cascade from the parameter row. The `report_benchmark` rows go in bounded chunks, each its own write statement, so a set with a long history does not hold the single writer connection for its entire cascade. The reports themselves survive the delete, so each affected report's metric count rollup is decremented rather than left overstating what the report holds.

## The one refusal

A benchmark's empty parameter set cannot be deleted. The empty set is structural: every benchmark is born with exactly one, and report ingest treats a missing empty set as data corruption rather than a set to mint, so this endpoint must not be able to manufacture that corruption. The request is refused with a conflict. Archiving the empty set stays allowed, because a later report revives it the same way it revives any other archived set.

## Tests

Integration tests cover the list with its archived filter and its pagination, the one get and its scoping to its own benchmark, create, create on a duplicate, create at the ceiling, the archive and unarchive round trip, the delete cascade down to the `report_benchmark`, metric, and series rows, the refusal on the empty set, and the permission failures on every write.

### DIFF
--- a/lib/api_projects/src/lib.rs
+++ b/lib/api_projects/src/lib.rs
@@ -18,6 +18,7 @@ pub mod jobs;
 mod keys;
 pub mod measures;
 pub mod metrics;
+pub mod parameters;
 pub mod perf;
 pub mod plots;
 pub mod projects;
@@ -142,6 +143,17 @@ impl bencher_endpoint::Registrar for Api {
         api_description.register(benchmarks::proj_benchmark_get)?;
         api_description.register(benchmarks::proj_benchmark_patch)?;
         api_description.register(benchmarks::proj_benchmark_delete)?;
+
+        // Parameters
+        if http_options {
+            api_description.register(parameters::proj_parameters_options)?;
+            api_description.register(parameters::proj_parameter_options)?;
+        }
+        api_description.register(parameters::proj_parameters_get)?;
+        api_description.register(parameters::proj_parameter_post)?;
+        api_description.register(parameters::proj_parameter_get)?;
+        api_description.register(parameters::proj_parameter_patch)?;
+        api_description.register(parameters::proj_parameter_delete)?;
 
         // Measures
         if http_options {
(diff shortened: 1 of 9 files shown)


epompeii:
Recreated from a stack-compatible branch; superseded by the follow-up pull request.

Issue #368, opened 2024-04-16
Create a way to track and test SQL query plan changes


epompeii:
There was a serious performance regression in https://github.com/bencherdev/bencher/issues/363. It was fixed and a follow up issue was created to help prevent it and similar performance regressions from happening in the future: https://github.com/bencherdev/bencher/issues/367.

However, this got me wondering if there was a better way to solve this than a pure runtime performance test. What would it look like if you were able to have a snapshot of the SQL database query plan and then check to see if that plan changed. That is "snapshot" testing for SQL query plans (not to be confused with using a snapshot of the database for testing). This would sort of be like instruction count based benchmarking but for databases. The query plan helps to indicate that there may be an issue with the runtime performance.

epompeii:
In order for this to work, spinning up a test database needs to be easy.
Something like pgmock seems like the right direction: https://github.com/stackframe-projects/pgmock

It would need to work across languages though, so a separate WASM runtime instead of a browser/Node would be necessary.

PR #1012, opened 2026-08-26
Query perf per variant and fetch every metric

epompeii:
The perf query answered per benchmark and returned one value per point. A benchmark now has variants, and a measure now carries several metrics, so it answers per variant and returns all of them. A variant is a benchmark with one concrete parameter set.

## The parameters filter

`parameters` is a new query parameter on `/v0/projects/{project}/perf`. It is a comma separated list of URL encoded parameter sets. A variant is queried when at least one element is a subset of its set, so the filter is an OR across its elements and an AND within one. An element that names one key matches every variant that pins that key to that value, however many other keys it also pins. The empty element matches every variant, and leaving the parameter off queries every variant too.

The filter resolves to parameter set row identifiers in Rust, over the benchmark's own sets, so what reaches SQL is an indexed lookup and never a JSON predicate. A value is its canonical form there, so `16`, `16.0`, and `1.6e1` are one filter and one match. If the filter excludes every variant of a benchmark, then that benchmark returns no lines. That is not an error, and the other benchmarks of the same query still return theirs.

## The fan out

A line is now keyed per variant. The line type carries the parameter set it plots, between the benchmark and the measure, because that is the order the dimensions run in: branch, testbed, benchmark, parameters, measure, metric. Two variants of one benchmark are two lines, not one line with the points interleaved. The lines come out in variant creation order, so the empty parameter set every benchmark is born with comes first.

## The response types

`JsonPerfMetrics` is now `JsonPerfLine`, and the point type that carries the `metrics` map takes the name `JsonPerfMetrics`. Each name now says what its type holds.

## The metrics map

Each point carries `metrics`, every metric the measure ingested, keyed by name. Each entry carries its value and `boundaries`, the thresholds that checked it, with the boundary each produced and any alert it raised. `boundaries` is a list from birth, because a later layer lets several thresholds check one metric. It is absent on a metric that nothing checked.

## The deprecated fields

They keep saying what they always said. `metric` is still the metric triple, rebuilt from the `value` row and its `lower_value` and `upper_value` siblings, and the singular `threshold`, `boundary`, and `alert` are still the ones that checked the `value` row. A project that only ever reported the metric triple gets back every field it got back before, unchanged, with the new fields added beside them. A test pins that field by field, including the float bytes.

`metric` becomes optional, following the report response. A measure that names no `value` is stored, billed, and queryable, so it is still a point on the line. It carries its metrics in `metrics` and leaves the triple absent. Nothing an older client can produce reaches that case. The plot leaves such a point out of the line it draws, because a measurement with no point estimate has nothing to place on the axis.

## The query

Reads move off the `metric_boundary` view onto the `metric` table. All the metrics of a variant for one measure sit together on the index over the report benchmark, the measure, and the name, so one bounded range read returns every one of them, where the view had to seek each conventional name separately and could only ever return those three.

The boundary, its threshold, its model, and its alert are chained flat with explicit `ON` clauses rather than nested inside the boundary join. SQLite cannot flatten a compound right operand of an outer join, so the nested form makes it scan the whole boundary table once per request, however narrow the query is. Tracking the plan itself is #368. Every access in the plan is an index seek or a bounded range read.

The perf image endpoint shares these internals and keeps its current behavior. It gains its own parameters filter in a later layer. No view or table is added.

The migration byte compatibility test drops its perf leg, because it captures responses on both sides of the metric migration, which only works for a reader that goes through the view. What the perf response owes older clients is pinned where that response lives.

## The bounds

A query returns at most the first 256 lines, in the order the dimensions run, and a line is one series on one branch. Only a line that comes back spends the budget, and at most 256 permutations run per request, counted before their SQL, so a sparse grid pays for what it plots and the work stays bounded either way. The limits are one family now: 8 parameter keys, 64 entries per dimension list, 256 lines, 256 permutations. When the line budget clips a benchmark mid fan out, the first variants in creation order return and the rest wait for a filter. Each dimension list is truncated to its first 64 entries at deserialization, and `parameters` is one of them.

The old cap was an index product that asymmetric grids walked through. A 2x2x6x21 query executed 456 permutations against a documented 255, and three skip paths never consulted it at all. A characterization suite pinned the response across permutations before any of this moved.

## The select and the window

The per row select drops the five dimension rows, 41 of its 89 columns, which were identical on every row of a permutation. The dimensions now resolve once per query through the same cache the variants use. SQLite's sorter carries the whole row, so engine time drops 1.6x at moderate row counts and 4.3x at 768,000 rows in the synthetic harness, with deserialization scaling the same way.

A query that names no start time now plots the last 4 weeks. That is the same `DEFAULT_REPORT_HISTORY` the console's default plot bootstraps with, the report comment's perf links carry, and the pinned plot form seeds, all 4 sites one value now, and the response echoes the window it applied. An explicit start time is honored as it stands, however old.

## The rename

A grid point is now called a variant, a metric is one named value a measure reported, and a threshold checks a metric rather than gating it. Gating is the CI decision an alert feeds. The sweep lands here, over this change and the files that predate it, so no retired term survives the merge.


lib/api_projects/src/perf/mod.rs
```diff
@@ -135,11 +183,17 @@ pub async fn get_inner(
...
+        parameters,
         measures,
         start_time,
         end_time,
     } = json_perf_query;
 
+    // A query that names no window still gets one, so an unwitting caller asks
+    // for the recent history rather than for everything ever reported. The
+    // response carries the window that was applied, not the one that was asked
+    // for, so the caller can see what it got.
```
epompeii:
```suggestion
```

lib/api_projects/src/perf/mod.rs
```diff
@@ -189,121 +520,225 @@ async fn perf_results(
...
-    let gt_max_permutations = permutations > MAX_PERMUTATIONS;
-    let mut results = Vec::with_capacity(permutations.min(MAX_PERMUTATIONS));
+    let mut results = Vec::with_capacity(permutations.min(MAX_LINES));
+    // What is left of the two budgets. A permutation spends one of the
+    // permutations by running at all, and one line per line it returns, so a
+    // grid of permutations that hold nothing is bounded by the first and costs
+    // the caller nothing of the second. The last permutation the line budget
+    // reaches is clipped mid fan out rather than dropped.
+    let mut permutations_left = MAX_PERMUTATIONS;
+    let mut lines_left = MAX_LINES;
```
epompeii:
```suggestion
    let mut permutations_remaining = MAX_PERMUTATIONS;
    let mut lines_remaining = MAX_LINES;
```

lib/api_projects/src/perf/mod.rs
```diff
@@ -189,121 +520,225 @@ async fn perf_results(
...
+struct Permutation {
+    branch_uuid: BranchUuid,
+    head_uuid: Option<HeadUuid>,
+    testbed_uuid: TestbedUuid,
+    spec_id: Option<SpecId>,
+    benchmark_uuid: BenchmarkUuid,
+    measure_uuid: MeasureUuid,
+    /// The variants to query, or every one of them when the query neither filters
+    /// nor clips the fan out.
+    parameter_ids: Option<Vec<ParameterId>>,
```
epompeii:
```suggestion
    benchmark_uuid: BenchmarkUuid,
    parameter_ids: Option<Vec<ParameterId>>,
    measure_uuid: MeasureUuid,
```

Ordering benchmark, parameters, measure

lib/api_projects/src/perf/mod.rs
```diff
@@ -189,121 +520,225 @@ async fn perf_results(
...
-    project: &QueryProject,
+fn perf_query(
+    project_id: ProjectId,
     branch_uuid: BranchUuid,
     head_uuid: Option<HeadUuid>,
     testbed_uuid: TestbedUuid,
     spec_id: Option<SpecId>,
     benchmark_uuid: BenchmarkUuid,
     measure_uuid: MeasureUuid,
+    parameter_ids: Option<Vec<ParameterId>>,
```
epompeii:
```suggestion
benchmark_uuid: BenchmarkUuid,
    parameter_ids: Option<Vec<ParameterId>>,
measure_uuid: MeasureUuid,
```

Consistent ordering

lib/api_projects/src/perf/mod.rs
```diff
@@ -396,171 +839,309 @@ async fn perf_query(
...
+                start_time: None,
+                end_time: None,
+            },
+        ))
+        .to_string();
+        assert!(
+            sql.contains("`report_benchmark`.`parameter_id` IN"),
+            "{sql}"
+        );
+    }
```
epompeii:
```suggestion
```

NO ASSERTING ON SQL OUTPUT!

lib/api_projects/src/metrics.rs
```diff
@@ -195,6 +194,20 @@ pub(super) type MetricQuery = (
...
+    if let Some((query_threshold, query_model, query_alert)) = tma {
+        let threshold =
+            Some(query_threshold.into_threshold_model_json_for_project(project, query_model));
+        let alert = query_alert.map(QueryAlert::into_perf_json);
+        (threshold, alert)
+    } else {
+        (None, None)
+    }
+}
+
```
epompeii: Why the hell did you put this in `lib/api_projects/src/metrics.rs`? It seemlingly has nothing to do with the Metrics API? WTF?

lib/bencher_json/src/project/perf.rs
```diff
@@ -148,11 +184,27 @@ impl TryFrom<JsonPerfQueryParams> for JsonPerfQuery {
...
+        testbeds.truncate(MAX_DIMENSION_ENTRIES);
+        let mut benchmarks: Vec<BenchmarkUuid> = from_urlencoded_list(&benchmarks)?;
+        benchmarks.truncate(MAX_DIMENSION_ENTRIES);
+        // An absent filter and an empty one both mean the same thing: no filter.
+        // Spelling it out here keeps the empty string out of the list reader, which
+        // rejects an empty element.
+        let parameters = match parameters.as_deref() {
+            Some(parameters) if !parameters.is_empty() => from_urlencoded_list(parameters)?,
+            _ => Vec::new(),
+        };
```
epompeii: Don't we want to truncate the parameters entries too? Not the sets themselves, but the list of sets? Limit that to `MAX_DIMENSION_ENTRIES` as well.

Also it feels like we should do this down below after the `specs` with the other sub dimensions (heads and specs) and keeping the branch, testbed, benchmark ordering.

lib/bencher_json/src/project/perf.rs
```diff
@@ -339,34 +406,48 @@ pub const HEADS: &str = "heads";
...
+/// One line of a perf query: one variant of one benchmark, on one branch, one
+/// testbed, and one measure.
+///
+/// The parameter set sits between the benchmark and the measure because a line is
+/// what a variant plots: two parameter sets of one benchmark are two lines, not
+/// one line with the points interleaved.
 #[typeshare::typeshare]
 #[derive(Debug, Clone, Deserialize, Serialize)]
 #[cfg_attr(feature = "schema", derive(JsonSchema))]
 pub struct JsonPerfMetrics {
```
epompeii: Should we rename this to `JsonPerfLine`?

lib/bencher_json/src/project/perf.rs
epompeii: Then rename this to `JsonPerfMetrics`, since this now has `metrics`.

PR #1017, opened 2026-08-26
Name the alert's variant and read boundaries off the base tables

epompeii:
## The field

`JsonAlert` gains `variant`, the one the alert fired on. It sits between `benchmark` and `metric`, because that is the order the dimensions run in: branch, testbed, benchmark, variant, measure, metric.

Two variants of one benchmark under one threshold used to raise two alerts that read identically: same benchmark, same measure, same threshold, nothing to tell them apart. Now the alert names its variant.

Every surface that returns an alert carries it:

- the alerts list, `GET /v0/projects/{project}/alerts`
- the alert detail, `GET /v0/projects/{project}/alerts/{alert}`, and its `PATCH` response
- the alerts a report response embeds

The perf response's `JsonPerfAlert` is unchanged. It is a slim reference, and the line it hangs off already names its variant.

## The reader moves

Both alert readers move off the `metric_boundary` view and onto the base tables. An alert names its boundary, the boundary names its metric row, and that row names the report it landed in and the variant it was measured under, so each hop is an identifier seek off the hop before it.

The metric triple is assembled from the `value` row the boundary was computed for plus its `lower_value` and `upper_value` siblings, which is the same assembly the metric endpoint does. That assembly is shared rather than written twice. Both readers carry the two siblings in their one query, as two left joins on `metric`, so an alert costs no bounds query of its own.

An alert whose boundary hangs off a row that is not a `value` row is refused with an error, not answered with numbers the alert does not name.

Alerts have a total order. The report response's embedded alerts and every sort of the alerts list end on the alert's identifier, so two variants of one benchmark that alert in one iteration come back in one order, and a page boundary cannot split or repeat them.

The move is what the next layer needs. The view carries at most one boundary per metric row, and a metric row is about to be allowed several, so a reader that reached its boundary through the view's left join would fan one alert out into one per boundary.

## No reader joins a boundary through the view

```
git grep -n 'view::metric_boundary'
```

One match, in `services/console/src/chunks/engineering/sqlite-performance-tuning/bonus-bug-rust-left-join.mdx`: a code sample quoted in a blog post, not a query. There is no other reference anywhere in the tree.

The view itself stays, and the migration that holds its column list unchanged stays pinned to it. Its Rust model and its Diesel table declaration are gone, because nothing reads them.

## Compatibility

Every field an alert carried before carries exactly what it carried before. `metric` included: it is still required, and it is still the triple built around the row the boundary was computed for.

The claim is stated as a fixture rather than asserted. A two variant history under one threshold is pinned key set and values both: the key set is the old one plus `variant`, and every old key holds the value the detector computes for that history. Four more fixtures pin the triple for a metric row with both bounds, either bound alone, and no bounds, on the alerts list, the alert detail, and the report response. The report response's embedded alerts and the alerts endpoint's alerts are pinned equal for the same report.

`cargo gen-types` regenerates the OpenAPI spec and the TypeScript types.


epompeii:

Rebased onto devel with the terminology sweep (grid point to variant, gate to check). Two independent reviews and an adjudication found no major findings in this layer.

Rebase changes:
- The alert test helper that lands two variants is now `ingest_two_variants`, because devel already has an `ingest_variants`.
- Two `report` fixture calls gained the `bmf_version` argument devel added to that helper.

One observation, not a defect: `QueryMetric::triple` adds one indexed lookup per alert, so the alerts list goes from one query per alert to two. It is bounded by the page size.

PR #1018, opened 2026-08-26
Check any metric of any variant, and let every matching threshold fire

epompeii:
A threshold checked one thing: the `value` metric of every variant of its measure.
A benchmark now reports one variant per distinct parameters and as many
metrics as its harness measured, and neither was addressable. A project that
wanted `p99` watched on the one configuration where it matters had nowhere to say
so.

## The model

A threshold gains two optional fields.

`metric` is the name it checks. A threshold that names none checks the conventional
`value` name: a threshold always checks exactly one name, and a bare one never checks
all of them.

`parameters` is a filter over variants: a list of partial parameters, an OR across
the list and a subset match within each entry. A filter names only the keys it cares
about, so `[{"size": 512}]` matches a variant that also pins `threads`. A
threshold with no filter checks every variant.

Both defaults are what every threshold already does, so no existing row moves and no
project's alert volume changes.

## Every matching threshold fires

There is no winner: not in checking, not in display. A variant that a bare
threshold and a filtered threshold both match earns a boundary from each and, on a
regression, an alert from each. Two thresholds asked for that row to be watched, so
two of them are told about it. This is deliberate and it is pinned by a test rather
than smoothed over.

## Identity, and the null semantics

Identity is the three dimensions plus the two new fields, under the null semantics
they carry. SQLite treats nulls as distinct in a unique index, so two bare
thresholds on one branch, testbed, and measure would no longer collide under a plain
unique key over the five columns, and an explicit `value` would sit beside an absent
one.

The key is declared over the effective values instead:

```sql
CREATE UNIQUE INDEX index_threshold_dimensions ON threshold(
    branch_id,
    testbed_id,
    COALESCE(parameters, x''),
    measure_id,
    COALESCE(metric, 'value')
);
```

`x''` is a value no stored filter can take, because a filter that matches everything
is stored as `NULL`.

The wire canonicalizes into those stored values. An explicit `value` and an absent
name are one threshold. An empty filter, a filter holding the empty entry, and an
absent filter are one threshold, because empty parameters are a subset of every
variant's. A filter has one spelling, because its entries sort by their RFC 8785
canonical bytes, duplicates collapse, and an entry another entry already covers is
dropped: `[{"a": 1}, {"a": 1.0}]` and `[{"a": 1}, {"a": 1, "b": 2}]` are both the one
entry `[{"a": 1}]`, because every variant the wider entry matches the narrower one
matches too. The empty entry rule is that rule's special case.
The wire accepts any order and any spelling; the response returns the canonical form.

## The sample keys on the name

The historical query behind detection filters on the threshold's metric name, so a
threshold on `p99` is tested against `p99` rows and never against the `value` rows
beside them. The per variant separation is unchanged: a threshold still samples
one variant's history and never the benchmark's variants pooled.

## The boundary key

`boundary` keys on `(metric_id, threshold_id)` rather than on `metric_id` alone,
because a metric row may now carry a boundary per threshold that checked it. Both
`threshold` and `boundary` are rebuilt, each with its unique keys built after the
copy rather than declared on the table, which is what keeps the rebuild at the cost
of the scan. `down.sql` restores both old shapes, and because the old schema has no
shape for a threshold that names a metric or checks only some variants, it deletes
every such threshold with its models, boundaries, and alerts rather than flattening
it into a bare one.

## Responses

`JsonAlert` gains `value`, the metric value the alert fired on, and its `metric`
triple becomes optional. The triple is a convention over the `value` name, so an
alert on any other name has none; every alert a threshold could raise before this
carries the triple exactly as it did, so no existing response changes. The checked
name is readable at `alert.threshold.metric`.

The base of this branch carried a `debug_assert_eq!` in `into_json_for_report`
holding that a checked row is always a point estimate. That is the invariant this
branch retires, so the assertion is removed by design rather than by accident, and
the comment in its place says so.

The deprecated singular `threshold`, `boundary`, and `alert` fields carry the bare
threshold's check and no other's, wherever they appear. That is precisely what a
caller from before named checking has always been shown: a row that only a named or
filtered threshold checks reports no check in them at all. Where a list of boundaries
is returned, it is ordered by the threshold UUID, which is creation order for a
UUIDv7 and deterministic for the UUIDv4 a threshold minted before the move to
UUIDv7. Nothing about that order is a ranking.

Every entry of that list names what its threshold checks: `JsonThresholdModel` gains
`parameters` and `metric`, both absent when the threshold takes the default, so a
reader looking at two boundaries on one row can tell the filtered one from the bare
one without a second request. The hand written `ParameterFilter` and `ParameterSet`
schemas now carry the caps they enforce, so a generated client can see them.

The in-report `thresholds.models` map is unchanged and still addresses the bare
threshold. A map that names a measure and a model says nothing about a name or a
variant, so it neither creates nor resets anything narrower. A threshold that
checks a name or only some variants is created through the thresholds endpoint.


epompeii:

Rebased onto devel with the terminology sweep (grid point to variant, gate to check, named value to metric). Two independent reviews and an adjudication found no major findings in this layer.

Rebase changes:
- Conflicts with devel's `boundary_json` and `deprecated_check` renames resolved: the bare threshold's singular fields are now carried as `bare_check`.
- Devel's boundary history query is now the hand-built `HistoryQuery` that walks the head newest first. This layer binds the threshold's metric name into it in place of `value`, leaving devel's SQL text unchanged, and devel's own query test builds its fixture as a bare `value` threshold.
- `Threshold::gates` is now `Threshold::checks`.
- Test fixtures pin an explicit perf `start_time`, because devel's four week default window postdates the 2024 fixture dates.
- `into_report_results_json` grew past the clippy line limit after the rebase, so its per row fold is extracted into `push_result_row`.

Cleanups applied from review: the `namedgate` fixture label, the `create_threshold_with_identity` test helper rename, and a narrowed comment in `metric_migration.rs`.

lib/api_projects/src/perf/mod.rs
```diff
@@ -725,13 +725,17 @@ fn perf_query(
...
                 (
                     schema::threshold::id,
                     schema::threshold::uuid,
                     schema::threshold::project_id,
-                    schema::threshold::measure_id,
                     schema::threshold::branch_id,
                     schema::threshold::testbed_id,
+                    schema::threshold::measure_id,
+                    schema::threshold::metric,
+                    schema::threshold::parameters,
```
epompeii: Preferred logical ordering:

```suggestion
                    schema::threshold::testbed_id,
                    schema::threshold::parameters,
                    schema::threshold::measure_id,
                    schema::threshold::metric,
```

lib/api_projects/src/thresholds.rs
```diff
@@ -307,13 +314,22 @@ pub async fn post_inner(
     let measure_id =
         QueryMeasure::from_name_id(auth_conn!(context), project_id, &json_threshold.measure)?.id;
 
-    // Create the new threshold
+    // Create the new threshold. What it checks beyond its dimensions is canonicalized
+    // here, so an explicit `value` and an absent metric are one threshold, and so are
+    // an empty filter and an absent one.
+    let identity = ThresholdIdentity::new(
+        json_threshold.metric.clone(),
+        json_threshold.parameters.clone(),
```
epompeii: Preferred  logical ordering:

```suggestion
        json_threshold.parameters.clone(),
        json_threshold.metric.clone(),
```

lib/bencher_json/src/project/parameter/jsonb.rs
```diff
@@ -111,6 +111,31 @@ impl Object {
...
+    }
+
+    /// The JSONB encoding of the array.
+    pub fn into_blob(self) -> Result<Vec<u8>, JsonbError> {
+        let mut blob = Vec::with_capacity(self.0.len().saturating_add(5));
+        push_element(&mut blob, ARRAY, &self.0)?;
+        Ok(blob)
+    }
+}
+
```
epompeii: Do we have any tests for this?

lib/bencher_json/src/project/parameter/mod.rs
```diff
@@ -184,6 +184,164 @@ pub enum ParametersError {
...
+            array: Some(Box::new(ArrayValidation {
+                items: Some(ParameterSet::json_schema(generator).into()),
+                ..Default::default()
+            })),
+            ..Default::default()
+        }
+        .into()
+    }
+}
+
```
epompeii: This along with its tests feels like they belong in their own file.

lib/bencher_json/src/project/threshold.rs
```diff
@@ -25,6 +25,16 @@ pub struct JsonNewThreshold {
...
+    /// The name of the metric this threshold checks.
+    /// If not set, the threshold checks the conventional `value` name.
+    /// A threshold always checks exactly one name.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub metric: Option<MetricName>,
+    /// The variants this threshold checks, as a parameters filter.
+    /// A variant matches when any entry in the filter is a subset of its parameters.
+    /// If not set, or set to an empty list, the threshold checks every variant.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub parameters: Option<ParameterFilter>,
```
epompeii: Preferred logical order:

```suggestion
    /// The variants this threshold checks, as a parameters filter.
    /// A variant matches when any entry in the filter is a subset of its parameters.
    /// If not set, or set to an empty list, the threshold checks every variant.
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub parameters: Option<ParameterFilter>,
   /// The UUID, slug, or name of the threshold measure.
   pub measure: MeasureNameId,
    /// The name of the metric this threshold checks.
    /// If not set, the threshold checks the conventional `value` name.
    /// A threshold always checks exactly one name.
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub metric: Option<MetricName>,
```

lib/bencher_json/src/project/threshold.rs
```diff
@@ -44,6 +54,14 @@ pub struct JsonThreshold {
...
     pub testbed: JsonTestbed,
     pub measure: JsonMeasure,
+    /// The name of the metric this threshold checks.
+    /// Absent when the threshold checks the conventional `value` name.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub metric: Option<MetricName>,
+    /// The variants this threshold checks, in canonical order.
+    /// Absent when the threshold checks every variant.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub parameters: Option<ParameterFilter>,
```
epompeii: Preferred logical order:

```suggestion
    /// The variants this threshold checks, in canonical order.
    /// Absent when the threshold checks every variant.
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub parameters: Option<ParameterFilter>,
    pub measure: JsonMeasure,
    /// The name of the metric this threshold checks.
    /// Absent when the threshold checks the conventional `value` name.
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub metric: Option<MetricName>,
```

lib/bencher_json/src/project/threshold.rs
```diff
@@ -59,6 +77,20 @@ pub struct JsonThresholdModel {
...
+    ///
+    /// A metric row may carry a boundary per threshold that checked it, and the
+    /// thresholds are put in the order they were created, oldest first, with the
+    /// UUID breaking a tie between two created in the same second. Nothing about the
+    /// list is a ranking: every threshold that checked the row is in it, and no reader
+    /// should read the first as the winner.
+    #[must_use]
+    pub fn boundary_order(&self) -> (i64, ThresholdUuid) {
+        (self.created.timestamp(), self.uuid)
+    }
```
epompeii: Didn't we move over to using UUIDv7? So shouldn't we just be able to use the UUID?

lib/bencher_schema/migrations/2026-08-26-130000_threshold_metric_parameters/up.sql
```diff
@@ -0,0 +1,171 @@
...
+-- copy has landed, as named indexes.
+CREATE TABLE up_threshold (
+    id INTEGER PRIMARY KEY NOT NULL,
+    uuid TEXT NOT NULL,
+    project_id INTEGER NOT NULL,
+    branch_id INTEGER NOT NULL,
+    testbed_id INTEGER NOT NULL,
+    measure_id INTEGER NOT NULL,
+    metric TEXT,
+    parameters BLOB,
```
epompeii: Preferred logical order:

```suggestion
    parameters BLOB,
    measure_id INTEGER NOT NULL,
    metric TEXT,
```

lib/bencher_schema/migrations/2026-08-26-130000_threshold_metric_parameters/up.sql
```diff
@@ -0,0 +1,171 @@
...
+-- and would let an explicit `value` sit beside an absent one. The index is declared
+-- over the effective values instead: NULL metric reads as `value` and NULL
+-- parameters reads as the empty blob, which is a value no stored filter can take
+-- because a filter that matches everything is stored as NULL.
+CREATE UNIQUE INDEX index_threshold_dimensions ON threshold(
+    branch_id,
+    testbed_id,
+    measure_id,
+    COALESCE(metric, 'value'),
+    COALESCE(parameters, x'')
```
epompeii: Preferred logical order:

```suggestion
    COALESCE(parameters, x''),
    measure_id,
    COALESCE(metric, 'value')
```

lib/bencher_schema/src/model/project/report/results/detector/threshold.rs
```diff
@@ -18,9 +22,17 @@ use crate::{
...
 pub struct Threshold {
     pub id: ThresholdId,
+    /// The name this threshold checks. A threshold that names none checks the
+    /// conventional `value` name, and a threshold always checks exactly one name.
+    pub metric: MetricName,
+    /// The variants this threshold checks. `None` checks every variant.
+    pub parameters: Option<ParameterFilter>,
+    /// When the threshold was created, which is what orders the candidates.
+    pub created: DateTime,
     pub model: ThresholdModel,
```
epompeii: Preferred logical ordering:

```suggestion
    /// The variants this threshold checks. `None` checks every variant.
    pub parameters: Option<ParameterFilter>,
    /// The name this threshold checks. A threshold that names none checks the
    /// conventional `value` name, and a threshold always checks exactly one name.
    pub metric: MetricName,
    pub model: ThresholdModel,
    /// When the threshold was created, which is what orders the candidates.
    pub created: DateTime,
```

lib/bencher_schema/src/model/project/report/results/mod.rs
```diff
@@ -61,7 +61,10 @@ pub struct ReportResults {
     pub benchmark_cache: HashMap<BenchmarkNameId, BenchmarkId>,
     pub variant_cache: HashMap<(BenchmarkId, ParameterSet), VariantId>,
     pub measure_cache: HashMap<MeasureNameId, MeasureId>,
-    pub detector_cache: HashMap<MeasureId, Option<Detector>>,
+    /// Every threshold of this report's (branch, testbed) that runs a model, grouped
+    /// by measure. Read once per report and matched in memory, so a variant that
+    /// several thresholds check costs no extra query.
+    pub thresholds: HashMap<MeasureId, Vec<Threshold>>,
```
epompeii:
```suggestion
    pub threshold_cache: HashMap<MeasureId, Vec<Threshold>>,
```

lib/bencher_schema/src/model/project/threshold/mod.rs
```diff
@@ -56,11 +56,73 @@ pub struct QueryThreshold {
     pub branch_id: BranchId,
     pub testbed_id: TestbedId,
     pub measure_id: MeasureId,
+    /// The name this threshold checks, when it is not the conventional `value` name.
+    pub metric: Option<MetricName>,
+    /// The variants this threshold checks, when it does not check every one.
+    pub parameters: Option<ParameterFilter>,
```
epompeii:
```suggestion
    /// The variants this threshold checks, when it does not check every one.
    pub parameters: Option<ParameterFilter>,
    /// The name this threshold checks, when it is not the conventional `value` name.
    pub metric: Option<MetricName>,
```

lib/bencher_schema/src/model/project/threshold/mod.rs
```diff
@@ -56,11 +56,73 @@ pub struct QueryThreshold {
...
+///
+/// Both halves are stored in their canonical form, which is the absence of a value
+/// for the default: the conventional `value` name is a `NULL` metric and a filter
+/// that matches every variant is `NULL` parameters. The wire accepts either
+/// spelling and this is what turns one into the other, so two spellings of one
+/// threshold are one row.
+#[derive(Debug, Clone, Default, PartialEq, Eq, Hash)]
+pub struct ThresholdIdentity {
+    pub metric: Option<MetricName>,
+    pub parameters: Option<ParameterFilter>,
```
epompeii:
```suggestion
    pub parameters: Option<ParameterFilter>,
    pub metric: Option<MetricName>,
```

lib/bencher_schema/src/model/project/threshold/mod.rs
```diff
@@ -56,11 +56,73 @@ pub struct QueryThreshold {
...
+/// What a threshold checks beyond its branch, testbed, and measure: one metric name
+/// and a filter over variants.
+///
+/// Both halves are stored in their canonical form, which is the absence of a value
+/// for the default: the conventional `value` name is a `NULL` metric and a filter
+/// that matches every variant is `NULL` parameters. The wire accepts either
+/// spelling and this is what turns one into the other, so two spellings of one
+/// threshold are one row.
+#[derive(Debug, Clone, Default, PartialEq, Eq, Hash)]
+pub struct ThresholdIdentity {
```
epompeii: I don't really like having this split from `ThresholdDimensions` like this. Do you ahve a good reason to do so if these are`Option`s anyway?

lib/bencher_schema/src/model/project/threshold/mod.rs
```diff
@@ -650,11 +790,19 @@ impl InsertThreshold {
...
+        // the threshold it addresses is the bare one: the `value` name of every
+        // variant. A threshold that checks a name or only some variants is addressed
+        // through the thresholds endpoint, so it is not what this updates and not
+        // what `reset` takes a model away from.
         let mut current_thresholds = schema::threshold::table
             .filter(schema::threshold::project_id.eq(project_id))
             .filter(schema::threshold::branch_id.eq(branch_id))
             .filter(schema::threshold::testbed_id.eq(testbed_id))
+            .filter(schema::threshold::metric.is_null())
+            .filter(schema::threshold::parameters.is_null())
```
epompeii:
```suggestion
            .filter(schema::threshold::parameters.is_null())
            .filter(schema::threshold::metric.is_null())
```

PR #1021, opened 2026-08-26
Pin the parameters filter on a plot

epompeii:
A pinned plot stores what the perf URL stores.

The other dimensions of a plot are entity references, so they live in join tables of UUIDs. The parameters dimension in query space is not a reference: it is a value predicate, a filter, so the plot stores the same canonical list of entries the perf query takes.

A UUID join table was considered and rejected. A pinned variant's full parameters used as a filter also match future supersets of themselves, so UUID pinning cannot faithfully drive the ruled query, and a filter column round-trips "pin the current perf view" exactly.

## The field

The `plot` table gains a nullable `parameters` column: the SQLite JSONB encoding of a JSON array of partial parameters, the same encoding the `threshold` column uses. `JsonPlot`, `JsonNewPlot`, and the plot patch shape gain `parameters` after `benchmarks` and before `measures`, the canonical dimension order.

NULL is match all, so every plot that predates this migration carries NULL and draws exactly what it drew before. A plot with no filter answers with the field absent, and the migration is a plain `ADD COLUMN`.

## Canonical semantics

The filter means exactly what the perf query's `parameters` param means: OR across the list, subset match within each entry. A variant matches when any entry of the list is a subset of its parameters, so a filter names only the keys it cares about and a variant that pins more keys still matches.

Canonicalization is `ParameterFilter`'s own, unchanged and unduplicated: each entry in its RFC 8785 canonical form, the list sorted by canonical bytes and deduplicated, at most eight entries. A list holding the empty entry and the empty list are both match all, and so is an absent field, so all three are one stored state: no filter at all.

On a patch, an absent `parameters` leaves the plot's filter alone. An explicit `null` and an empty list both clear it back to every variant, which are two spellings of one filter and so land on one stored value.

The console will pass the filter through when it builds the perf query. No console change here.


epompeii:

Rebased onto devel with no source conflicts. Two independent reviews and an adjudication found no major defect introduced by the rebase.

Cleanup applied from review: the `plot_parameters` migration now recreates the table so `parameters` sits after `window`, rather than `ADD COLUMN` placing it after `modified`. `schema.rs`, `QueryPlot`, and `InsertPlot` follow the same order.

### Follow-up: a pinned plot's filter never reaches the console

A plot's `parameters` filter is stored and returned, but `PinnedFrame.tsx` builds the perf query from branches, testbeds, benchmarks, and measures only, so a plot pinned with a filter through the API or `bencher plot create`/`update` draws every variant in the console. `specs` is missing from the same query already. This is console work for the layer that renders variants: pass the plot's `parameters` into the pinned frame's perf query.

PR #1028, opened 2026-09-15
Create a branch head and its start point versions in one transaction

epompeii:
A new branch head was committed in one write transaction, and its start point versions were copied into it afterward in a separate write. A report uploaded to the branch in that window found the head empty and numbered its version 0, even though its version id was newer than every version the copy then added with higher numbers. The head's version numbers stopped rising with version id, so history ordered by version number treated the newest report as the oldest.

Branch creation with a start point and head replacement (a start point change or reset) now read the start point's version ids first and insert them inside the same transaction that creates the head. Versions added to the start point afterward carry larger numbers and are excluded by the existing number filter, so reading early is safe. The `max_versions` limit and ordering are unchanged. If the copy fails, the head creation rolls back instead of leaving a head without its versions. Threshold cloning still runs after the commit.

The separate clone step is removed, along with its redundant re-read of the start point version.

Testing: the race needs concurrent requests, so the tests pin the atomicity the fix provides. Two API tests inject a failure into `head_version` inserts with a SQLite trigger, then assert that creating a branch from a start point leaves no branch behind and that resetting a branch's start point keeps its old head. Both fail on the previous code. Each test then removes the trigger, retries, and checks that the new head holds exactly the start point's versions.


epompeii:
Recreated as https://github.com/bencherdev/bencher/pull/1031. The branch was recreated as `u/ep/boundary/head-versions-atomic` so that CI runs on the stack.

PR #1029, opened 2026-09-15
Renumber versions atomically when a report delete removes its version

epompeii:
When deleting a report removes the last report on a version, that version is deleted and every later version in the heads that contain it is decremented. That work previously ran outside any transaction: the remaining reports check used a read connection, each later version was written back one statement at a time with a number read earlier, and a failed update was logged and skipped before the version was deleted anyway. A failed update could leave duplicate numbers, two concurrent deletes could write stale numbers, and a concurrent upload could attach a report to the version between the check and the delete.

Now, after the report row is deleted, the remaining reports check, the renumber, and the version delete run in one write transaction. The renumber is a single `UPDATE version SET number = number - 1` over the later versions in the heads that contain the deleted version, so each version decrements once from its own current value. Any error rolls the transaction back and fails the request with a conflict. The set of versions that get decremented is unchanged. The per-version map and loop, and the now unused `VersionNumber::decrement`, are removed. If the version is already gone (a concurrent delete of its last other report got there first), the delete succeeds without renumbering.

Tests in `lib/api_projects/tests/reports.rs`:
- `reports_delete_renumbers_shared_versions_once`: a version shared by two heads that also share two later versions; deleting its report removes it, decrements each later version exactly once, keeps numbers strictly increasing in both heads, and leaves a version in an unrelated head untouched. This already passed before the change and pins the behavior.
- `reports_delete_version_failure_rolls_back_renumber`: a trigger makes the version delete fail after the renumber has run; the delete returns a conflict and the renumber rolls back, leaving every version number as it was. Before the change the request also returned a conflict, but the later versions stayed decremented.

The concurrent delete and concurrent upload races have no practical deterministic test; the single write transaction is what closes them.

Stacked on https://github.com/bencherdev/bencher/pull/1028.


epompeii:
Recreated as https://github.com/bencherdev/bencher/pull/1032. The branch was recreated as `u/ep/boundary/report-delete-renumber` so that CI runs on the stack.

PR #1030, opened 2026-09-15
Walk the head newest first in the boundary history query

epompeii:
Each threshold check during a report upload loads the variant's recent history: the newest `value` metrics for one benchmark variant and measure on one head and testbed, capped at the model's maximum sample size, with optional spec and window filters. Without planner statistics, SQLite started that query from the variant's whole history across every branch, so a short head on a deep history read far more rows than it returned.

## What changed

- `metrics_data` is now a `sql_query` with typed binds. A pinned `CROSS JOIN` order starts at `head_version` for the head, then joins the version's reports on the testbed, their results for the variant, and the metric for the measure.
- It orders by `head_version.version_id DESC, report.start_time DESC, report_benchmark.iteration DESC`. The head's `(head_id, version_id)` unique index supplies the version order, so with a sample size limit the walk stops once the limit is met instead of sorting the whole history.
- Ordering by version id matches ordering by version number because a head's version numbers rise with its version ids.
- The spec filter, window, and limit are appended only when present. There are no new indexes, columns, or migrations, and `report_benchmark` is pinned to its `(benchmark_id, report_id)` index so each report seeks the variant's rows directly instead of scanning all of its results.
- The walk stops early only once the sample size limit is met. A check with no limit, or with fewer matching samples on the head (or in its window) than its limit, walks the whole head on every upload, as the previous query also read that whole history. That cost is accepted to keep the query simple.

## Testing

- `metrics_data_newest_version_first` seeds a branch cloned from a start point with its own newer version, a second testbed, a spec on some reports, reports inside and outside a window, several reports on one version with equal start times and interleaved iterations, and noise rows for another parameter, another benchmark, another measure, and a bound metric name. It asserts the exact values in the exact order for the spec, window, and limit combinations on both heads. The test passed against the previous query before the rewrite, and still passes.
- `metrics_data_plan_walks_head_newest_first` checks `EXPLAIN QUERY PLAN` for all eight query shapes, with no statistics and again after `ANALYZE`. The plan must search `head_version` through `sqlite_autoindex_head_version_1` and must not use a temp B-tree for the version term of the `ORDER BY`. Ordering by `version.number` instead fails this test.

Stacked on https://github.com/bencherdev/bencher/pull/1029.


epompeii:
Recreated as https://github.com/bencherdev/bencher/pull/1033. The branch was recreated as `u/ep/boundary/newest-first` so that CI runs on the stack.

